PR #2, opened 2023-07-30
Fix early loading errors

cpw:
Fixes unhandled crashes when required files are not in proper state.

### DIFF
--- a/loader/src/main/java/net/minecraftforge/fml/loading/FMLPaths.java
+++ b/loader/src/main/java/net/minecraftforge/fml/loading/FMLPaths.java
@@ -69,8 +69,6 @@ public enum FMLPaths
                     throw new UncheckedIOException(ioe);
                 }
             }
-            if (path.isDirectory && Files.exists(path.absolutePath) && !Files.isDirectory(path.absolutePath)) {
-            }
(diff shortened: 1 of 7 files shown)


TheCurle (approved):
Better error handling is always a good thing, and fixes an issue I encountered today where the game crashes without detail if the mods folder could not be created for whatever reason.

Will likely merge if nothing comes up by tomorrow.

loader/src/main/java/net/minecraftforge/fml/loading/FMLPaths.java
```diff
@@ -63,10 +65,12 @@ public static void loadAbsolutePaths(Path rootPath)
             {
                 try {
                    Files.createDirectories(path.absolutePath);
-                } catch (IOException e) {
-                   throw new RuntimeException(e);
+                } catch (IOException ioe) {
+                    throw new UncheckedIOException(ioe);
                 }
             }
+            if (path.isDirectory && Files.exists(path.absolutePath) && !Files.isDirectory(path.absolutePath)) {
```
ThatGravyBoat, line 72: Is this `if` statement meant to be wrapping something? because right now it's just an empty body.

PR #3, opened 2023-08-02
Fix use of OS-dependent `File.separatorChar` in `FileUtils`

sciwhiz12:
This PR removes the use of the OS-dependent `File.separatorChar` in `FileUtils`, and substituting it with the use of the JDK's `Path` interface (which properly recognizes both separator slashes, even when mixed).  

This PR blocks the merge of neoforged/NeoForge#68, which in turn blocks the resolution of neoforged/NeoForge#38. See https://github.com/neoforged/NeoForge/pull/68#issuecomment-1662480101 for more details on what this resolves.


marchermans (approved):

cpw:
How did this survive for so many years?

PR #7, opened 2023-08-08
Allow mod constructors to receive useful arguments

Technici4n:
This allows mod to optionally receive their mod bus and mod container via constructor arguments.

As a next step, I imagine methods like `registerConfig` and `registerExtensionPoint` will be moved from `ModLoadingContext` to `ModContainer`.

Tested by publishing to maven local and loading it in forgedev. The errors also work as expected:
image
image

### DIFF
--- a/languages/java/src/main/java/net/minecraftforge/fml/javafmlmod/FMLModContainer.java
+++ b/languages/java/src/main/java/net/minecraftforge/fml/javafmlmod/FMLModContainer.java
@@ -76,7 +76,8 @@ public class FMLModContainer extends ModContainer
                 // Otherwise look for constructor that can accept more arguments
                 Map<Class<?>, Object> allowedConstructorArgs = Map.of(
                         IEventBus.class, eventBus,
-                        ModContainer.class, this);
+                        ModContainer.class, this,
+                        FMLModContainer.class, this);
 
                 constructorsLoop: for (var constructor : modClass.getDeclaredConstructors()) {
                     var parameterTypes = constructor.getParameterTypes();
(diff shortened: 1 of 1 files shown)


languages/java/src/main/java/net/minecraftforge/fml/javafmlmod/FMLModContainer.java
```diff
@@ -65,7 +69,44 @@ private void constructMod()
...
             LOGGER.trace(LOADING, "Loading mod instance {} of type {}", getModId(), modClass.getName());
-            this.modInstance = modClass.getDeclaredConstructor().newInstance();
+            try {
+                // Try noargs constructor first
+                this.modInstance = modClass.getDeclaredConstructor().newInstance();
+            } catch (NoSuchMethodException ignored) {
+                // Otherwise look for constructor that can accept more arguments
+                Map<Class<?>, Object> allowedConstructorArgs = Map.of(
+                        IEventBus.class, eventBus,
+                        ModContainer.class, this);
```
Technici4n, line 79: Question asked by @PepperCode1: should it be `FMLModContainer` instead?

Su5eD: No, `FMLModContainer` is specific to the java language loader. Other languages will provider their own implementation.

Technici4n: Hmmm, I'm pretty sure this code is java language loader-specific, so it should be fine here?

Su5eD: Sorry, thought this was fmlloader code. Yes in this case we can surely provide the implementation type.

Technici4n: Added `FMLModContainer` - mods can choose which one they want.

Technici4n:
Enforcing a single constructor would be better, but that would be a potential breaking change.

Matyrobbrt (approved):

Su5eD (approved):

cpw (approved):
OK this seems reasonable, yes.

PR #8, opened 2023-08-09
Correctly support EventPriority for non-parallel mod bus events

Technici4n:
Depends on neoforged/Bus#12.

Needs proper testing, which I haven't done yet.

Note that this is a breaking change, so we have to wait until 1.20.2.


Technici4n:
Question: what to do with `runEventGenerator`.

Technici4n:
Superseded by #27.

PR #14, opened 2023-08-20
Don't scan assets and data folders in the mod jars for .class files

Technici4n:
Depends on https://github.com/McModLauncher/securejarhandler/pull/43.


Technici4n:
Replaced by #38.

PR #18, opened 2023-08-22
Get the cause of InvocationTargetException when a constructor throws

### LINKED ISSUE #15: Fix java.lang.reflect.InvocationTargetException: null
It would be much better to get the actual exception instead of the exception reflection throws when a method call results in an exception.

Jab125:
Closes #15

### DIFF
--- a/languages/java/src/main/java/net/minecraftforge/fml/javafmlmod/FMLModContainer.java
+++ b/languages/java/src/main/java/net/minecraftforge/fml/javafmlmod/FMLModContainer.java
@@ -113,7 +113,7 @@ public class FMLModContainer extends ModContainer
         }
         catch (Throwable e)
         {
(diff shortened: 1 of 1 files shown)


languages/java/src/main/java/net/minecraftforge/fml/javafmlmod/FMLModContainer.java
```diff
@@ -70,6 +71,7 @@ private void constructMod()
         }
         catch (Throwable e)
         {
+            if (e instanceof InvocationTargetException) e = e.getCause();
```
Technici4n, line 74: might be nice to add a comment above to explain that this is done for the error screen

Technici4n:
Note: please resolve the conflict. :wink:

Technici4n (approved):

PR #19, opened 2023-08-23
Explicitly pass null filter when creating SecureJar/UnionFS

embeddedt:
`UnionFileSystem.testFilter` contains an optimization intended for the case where no path filter is set:

https://github.com/McModLauncher/securejarhandler/blob/1dedb0655d9f4b8cf310ea7095b38c7a978a42ab/src/main/java/cpw/mods/niofs/union/UnionFileSystem.java#L397

However, all of the code paths creating commonly-used UnionFS instances in FML are passing non-null `(a, b) -> true` lambdas, defeating this optimization. I have frequently seen `testFilter` appearing as a hotspot in my profiling while working on ModernFix.

The solution is quite simple: start passing `null` instead of a lambda. I've gone through the codebase and replaced all the lambdas matching this pattern.

Would appreciate review from someone familiar with this part of the codebase.

### DIFF
--- a/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/AbstractModProvider.java
+++ b/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/AbstractModProvider.java
@@ -38,7 +38,7 @@ public abstract class AbstractModProvider implements IModProvider
         var sj = SecureJar.from(
                 Manifest::new,
                 jar -> jar.moduleDataProvider().findFile(MODS_TOML).isPresent() ? mjm : JarMetadata.from(jar, path),
-                (root, p) -> true,
+                null,
                 path
         );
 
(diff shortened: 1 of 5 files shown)


loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/AbstractModProvider.java
```diff
@@ -38,7 +38,7 @@ protected IModLocator.ModFileOrException createMod(Path... path) {
         var sj = SecureJar.from(
                 Manifest::new,
                 jar -> jar.moduleDataProvider().findFile(MODS_TOML).isPresent() ? mjm : JarMetadata.from(jar, path),
-                (root, p) -> true,
+                null,
```
embeddedt, line 41: I've tested this specific class' change in production using my Java agent and an equivalent ASM transformer. The other changes have not been tested yet.

PR #21, opened 2023-09-14
Add implementations of IModFileInfo#showAsDataPack for data pack grouping

lukebemish:
NeoForged/neoforge#138, in addition to changing how resource grouping works, sets up a system to group mod datapacks in addition to mod resource packs. This is useful for a number of reasons - largest on the list is that it is a simple way of ensuring that mod datapacks are sorted in mod loading order, not "order I installed the mod" order.

This PR implements a new API that that PR will need - namely, a flag in mod file info on whether mods should show a separate data pack, to go along with the equivalent one that already exists for resource packs. It goes along with NeoForged/NeoForgeSPI#1, which adds the actual API, which this PR then implements; as such, it's not going to build without that PR


lukebemish:
https://github.com/neoforged/NeoForge/pull/138 is now superseded by https://github.com/neoforged/NeoForge/pull/367, but this PR is still needed for the latter

Technici4n (approved):

PR #23, opened 2023-09-29
Require side to be declared in mods.toml

0xilly:
After some discussion in the [discord](https://discord.com/channels/313125603924639766/313125603924639766/1157126571807752252) about how easy it is for users to drop a server side on the client and crash the server. To make the crash less opaque I propose we require modders to declare the side their mod is on during construction.

For this PR to work this [patch](https://github.com/neoforged/NeoForgeSPI/pull/2) for NeoForgeSPI needs to be merged

Some additional things for discussion for another PR, I do think we can do a better job collecting and displaying info to the end user server side there's some UX work that could be done.

### DIFF
--- a/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModInfo.java
+++ b/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModInfo.java
@@ -38,6 +38,7 @@ public class ModInfo implements IModInfo, IConfigurable
     private final ArtifactVersion version;
     private final String displayName;
     private final String description;
+    private final DependencySide side;
     private final Optional<String> logoFile;
     private final boolean logoBlur;
     private final Optional<URL> updateJSONURL;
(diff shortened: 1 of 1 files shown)


lukebemish:
Are there examples of mods which truly shouldn't be installed on the server or client? This seems like it puts a burden on modders who are doing stuff right, who now need to declare a side, just to benefit mods that should really be doing side checks to ensure they're on the right side anyways. Unless I've misunderstood and it's not mandatory in the mods.toml?

gigaherz:
I'm not convinced by this. 

While I can see the benefit in flagging a mod as "needs client, it won't work in dedicated server so don't bother trying to load it", I have encountered many times people forget that multiplayer exists and make a single player only mod that still fails to work in multiplayer, and I have seen many times people that forget the integrated server and LAN mode exist, and make a "server only" mod that would actually be useful (and works) in LAN and singleplayer, but flagging it as "dedicated server only, don't load in clients" would prevent using it for no reason.

Also as a second consideration, without any additional documentation people are likely to misinterpret what "side" does: they will use "client" thinking it won't load in the integrated server, or they will think that "client" means it will magically work in multiplayer.

Matyrobbrt:
I don't consider this the correct solution, for a few reasons.

First of all, *dedicated server*-only mods practically never exist. A lot of mods tweaking behaviour server-side could just as well do it on LAN worlds too. This would incentivise people to mark a mod as `Dist.DEDICATED_SERVER`, forgetting or otherwise ignoring LAN servers. 
As for client-only mods, I believe that side-only entrypoints are by far the better solution. It will go a longer way preventing client-only classes to be loaded on servers, while also being useful to any mod, as it would not prevent the loading of the entire mod.

Mgazul:
Just require side in the mod's main class

PR #26, opened 2023-10-04
Normalize CRLF mod description line endings to use LF instead

Su5eD:
### The issue

Minecraft can't properly display CRLF line endings in its GUIs. This leads to them being shown as characters in mod descriptions on the NeoForge mod list screen, as described in the following issue

- neoforged/NeoForge#104

#### Expected behavior

Line ending characters should not be visible by any means.
 
### The solution

When parsing mod descriptions, we replace all CRLF line endings with LF. By eliminating the root cause of this issue in FML, we ensure CRLFs will not show up anywhere in game, be it the neoforge mod list screen or other mod GUIs.

Fixes neoforged/NeoForge#104

| Before | After |
|--------|--------|
| broken crlf| fancy lf|

### DIFF
--- a/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModInfo.java
+++ b/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModInfo.java
@@ -80,6 +80,8 @@ public class ModInfo implements IModInfo, IConfigurable
         this.displayName = config.<String>getConfigElement("displayName")
                 .orElse(this.modId);
         this.description = config.<String>getConfigElement("description")
+                // Normalize CRLF line endings to use LF instead
+                .map(desc -> desc.replace("\r\n", "\n"))
                 .orElse("MISSING DESCRIPTION");
         this.logoFile = Optional.ofNullable(config.<String>getConfigElement("logoFile")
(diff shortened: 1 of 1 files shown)


Technici4n (approved):

TelepathicGrunt:
Would this then resolve this report?
https://github.com/neoforged/NeoForge/issues/104

PaintNinja:
I think this should also account for macOS which emits a `\r` without a `\n`

Technici4n:
> I think this should also account for macOS which emits a \r without a \n

Really? Source?

Su5eD:
> Your memory is from the good old times though: Mac OS X, as POSIX-compliant Unix uses the typical Unix LF.
> CR is a relict from the "classic" Mac OS, it's not used anymore.
> https://superuser.com/a/439443

Well then I guess it doesn't...

sciwhiz12:
Note that the TOML spec itself (both [v1.0.0](https://toml.io/en/v1.0.0#spec) and [v0.5.0](https://toml.io/en/v0.5.0#spec), where NightConfig is stated to be compliant up to the latter) only recognizes CR and CRLF as valid newline indicators, which aligns with Su5eD's finding (as shared via the Discord server) where the TOML parser rejects CR line endings. Therefore trying to account for the ancient CR line ending is moot.

PR #38, opened 2023-11-12
Apply SJH updates and remove deprecated members

Technici4n:
Depends on https://github.com/McModLauncher/securejarhandler/pull/47.

The `IModFileInfo` creation is still quite messy, but that will be for another time. :P

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/AbstractModProvider.java
+++ b/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/AbstractModProvider.java
@@ -6,6 +6,7 @@
 package net.neoforged.fml.loading.moddiscovery;
 
 import com.mojang.logging.LogUtils;
+import cpw.mods.jarhandling.JarContentsBuilder;
 import cpw.mods.jarhandling.JarMetadata;
 import cpw.mods.jarhandling.SecureJar;
 import net.neoforged.fml.loading.LogMarkers;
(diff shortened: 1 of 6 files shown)


loader/src/main/java/net/neoforged/fml/loading/moddiscovery/AbstractModProvider.java
```diff
@@ -34,30 +35,28 @@ public abstract class AbstractModProvider implements IModProvider
...
-        var mjm = new ModJarMetadata();
-        var sj = SecureJar.from(
-                Manifest::new,
-                jar -> jar.moduleDataProvider().findFile(MODS_TOML).isPresent() ? mjm : JarMetadata.from(jar, path),
-                null,
-                path
-        );
+        var jarContents = new JarContentsBuilder()
+                .paths(path)
+                .ignoreRootPackages("assets", "data")
```
Technici4n, line 40: Note: this skips scanning the `assets` and `data` folders of mods for classes... :)

loader/src/main/java/net/neoforged/fml/loading/ModDirTransformerDiscoverer.java
```diff
@@ -93,10 +95,10 @@ private static void visitFile(Path path) {
         if (!path.toString().endsWith(".jar")) return;
         if (LamdbaExceptionUtils.uncheck(() -> Files.size(path)) == 0) return;
 
-        SecureJar jar = SecureJar.from(path);
-        jar.moduleDataProvider().descriptor().provides().stream()
-            .map(ModuleDescriptor.Provides::service)
-            .filter(SERVICES::contains)
+        JarMetadata metadata = JarMetadata.from(new JarContentsBuilder().paths(path).build());
+        metadata.providers().stream()
```
Technici4n, line 99: This greatly speeds up scanning for providers because we are not computing the packages anymore.

loader/src/main/java/net/neoforged/fml/loading/ModDirTransformerDiscoverer.java
```diff
@@ -82,7 +84,7 @@ private static void scan(final Path gameDirectory) {
             return;
         }
         try (var walk = Files.walk(modsDir, 1)){
-            walk.forEach(ModDirTransformerDiscoverer::visitFile);
+            walk.parallel().forEach(ModDirTransformerDiscoverer::visitFile);
```
Technici4n, line 87: Walk in parallel to open file systems in parallel, should save some time too based on ATM9 profiles.

embeddedt: Does this actually work (i.e. does `Files.walk`'s stream actually support parallelism, or is it a no-op)? Usually I like to collect into a list first and then parallel stream that.

Technici4n: Apparently this does not work because of a JDK bug that was only fixed in Java 19: https://bugs.openjdk.org/browse/JDK-8280915. Thanks for noticing...

Technici4n: Fixed in 7a0fecd.

loader/src/main/java/net/neoforged/fml/loading/ModDirTransformerDiscoverer.java
```diff
@@ -82,8 +82,11 @@ private static void scan(final Path gameDirectory) {
             // Skip if the mods dir doesn't exist yet.
             return;
         }
-        try (var walk = Files.walk(modsDir, 1)){
-            walk
+        try (var walk = Files.walk(modsDir, 1)) {
+            // Collect to list first, and then parallel stream it.
+            // Before JDK 19, Files.walk streams are not parallelized efficiently for small numbers of elements.
+            // See https://bugs.openjdk.org/browse/JDK-8280915.
+            walk.toList().stream()
```
embeddedt, line 89: Why not `parallelStream` instead of `stream().parallel()`? They're likely equivalent but it feels like a code smell.

Technici4n: Just because I didn't know it existed. :P There doesn't seem to be anything justifying it though in the code, but whatever.

loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ExplodedDirectoryLocator.java
```diff
@@ -31,12 +33,17 @@ public record ExplodedMod(String modid, List<Path> paths) {}
...
     public List<IModLocator.ModFileOrException> scanMods() {
-        explodedMods.forEach(explodedMod ->
-                ModJarMetadata.buildFile(this,
-                        jar->jar.moduleDataProvider().findFile("/META-INF/mods.toml").isPresent(),
-                        null,
-                        explodedMod.paths().toArray(Path[]::new))
-                .ifPresentOrElse(f->mods.put(explodedMod, f), () -> LOGGER.warn(LogMarkers.LOADING, "Failed to find exploded resource mods.toml in directory {}", explodedMod.paths().get(0).toString())));
+        explodedMods.forEach(explodedMod -> {
+            var jarContents = new JarContentsBuilder().paths(explodedMod.paths().toArray(Path[]::new)).build();
+            if (jarContents.findFile("/META-INF/mods.toml").isPresent()) {
```
Matyrobbrt, line 38: Any reason this path has a leading slash?

Technici4n: That's just copy/pasted from the deprecated methods :P

PR #40, opened 2023-11-19
Fixing and cleaning up configs

dima-dencep:
The things that keep me from sleeping well. Commits are documented (if possible),

Perhaps after merge this pr I will make an additional one for neoforge, however right now it is not required and does not make any breaking changes (Probably)


loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
```diff
@@ -23,10 +23,12 @@
 public class FMLConfig
 {
     public enum ConfigValue {
+        DISABLE_CONFIG_WATCHER("disableConfigWatcher", Boolean.FALSE, "Disables File Watcher. Used to automatically update config if its file has been modified."),
```
Technici4n, line 26: What is the point of disabling the config watcher? If it is to fix a bug, it might already be fixed in upstream NightConfig. Or at least it was reported to TheElectronWill some time ago and he said a fix was in the pipeline?

core/src/main/java/net/neoforged/fml/config/ModConfig.java
```diff
@@ -24,17 +23,20 @@ public class ModConfig
     private final ModContainer container;
     private final ConfigFileTypeHandler configHandler;
     private CommentedConfig configData;
-    private Callable<Void> saveHandler;
 
-    public ModConfig(final Type type, final IConfigSpec<?> spec, final ModContainer container, final String fileName) {
+    public ModConfig(final Type type, final IConfigSpec<?> spec, final ModContainer container, final String fileName, ConfigFileTypeHandler configHandler) {
```
Technici4n, line 27: Why a custom handler? Is this to implement a custom format? (Not sure we want that?)

core/src/main/java/net/neoforged/fml/config/ModConfig.java
```diff
@@ -73,7 +75,7 @@ void setConfigData(final CommentedConfig configData) {
         this.spec.acceptConfig(this.configData);
     }
 
-    void fireEvent(final IConfigEvent configEvent) {
+    public void fireEvent(final IConfigEvent configEvent) {
```
Technici4n, line 78: We should get rid of `dispatchConfigEvent` and the `configHandler`, and instead just fire via `acceptEvent` or the event bus directly.

core/src/main/java/net/neoforged/fml/config/ConfigFileTypeHandler.java
```diff
@@ -31,38 +31,56 @@ public class ConfigFileTypeHandler {
...
+                    Files.delete(configData.getNioPath());
+
+                    configData.load();
+                }
+                catch (Throwable t)
+                {
+                    ex.addSuppressed(t);
+
+                    throw new ConfigLoadingException(c, ex);
+                }
```
Technici4n, line 61: What is this for?

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
```diff
@@ -23,10 +23,12 @@
...
         EARLY_WINDOW_CONTROL("earlyWindowControl", Boolean.TRUE, "Should we control the window. Disabling this disables new GL features and can be bad for mods that rely on them."),
         MAX_THREADS("maxThreads", -1, "Max threads for early initialization parallelism,  -1 is based on processor count", FMLConfig::maxThreads),
         VERSION_CHECK("versionCheck", Boolean.TRUE, "Enable NeoForge global version checking"),
         DEFAULT_CONFIG_PATH("defaultConfigPath", "defaultconfigs", "Default config path for servers"),
+        CONFIG_BACKUPS_COUNT("configBackupsCount", 5, "Sets the number of backups for configs."),
```
Technici4n, line 31: It was a TODO in the code, yet configuring a magic number like that feels kind of wrong?

Technici4n (changes requested):
A few comments since I'm not too familiar with the config code.

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
dima-dencep: Basically, there's no point as such, but I'd like to be able to do so

core/src/main/java/net/neoforged/fml/config/ConfigFileTypeHandler.java
dima-dencep: For something like this https://github.com/mezz/JustEnoughItems/issues/3317

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
dima-dencep: Maybe, but it's a good option (in my opinion).

core/src/main/java/net/neoforged/fml/config/ModConfig.java
dima-dencep: I had a choice between deleting the `configHandler` field or creating it via constructor, which is what I chose, but yes, custom formats are useless since toml is fine

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
embeddedt: FWIW, I'm in favor of this change - there are outstanding issues with the watcher on our end that are not addressed, such as the `Loading` & `Reloading` events being able to fire simultaneously, and the whole concept being quite broken with editors that don't save atomically. So I think giving users the option to disable the watcher is a good stopgap until someone has time to properly rewrite that logic.

core/src/main/java/net/neoforged/fml/config/ConfigFileTypeHandler.java
```diff
@@ -31,38 +31,56 @@ public class ConfigFileTypeHandler {
...
             final Path configPath = configBasePath.resolve(c.getFileName());
-            final CommentedFileConfig configData = CommentedFileConfig.builder(configPath).sync().
-                    preserveInsertionOrder().
-                    autosave().
-                    onFileNotFound((newfile, configFormat)-> setupConfigFile(c, newfile, configFormat)).
-                    writingMode(WritingMode.REPLACE).
-                    build();
-            LOGGER.debug(CONFIG, "Built TOML config for {}", configPath.toString());
+            final CommentedFileConfig configData = CommentedFileConfig.builder(configPath)
+                    .sync()
```
marchermans, line 35: Why was a sync added here?

core/src/main/java/net/neoforged/fml/config/ModConfig.java
marchermans: I don't support this change.
Fml solely supports toml configs and this opens the door to others, which fractures the eco system for no good reason. Please revert. The original code with setting the field in the constructor is fit for purpose.

Hell I would even argue the other way around and remove the field entirely

core/src/main/java/net/neoforged/fml/config/ModConfig.java
marchermans: Yes

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
marchermans: This does not really solve anything. The reason for the hardcode of the backups is that the fml config itself might fail to load and you then have no idea about the backup count the user is requesting.

I think for stability reasons this value should be hardcoded.

marchermans (changes requested):

core/src/main/java/net/neoforged/fml/config/ConfigFileTypeHandler.java
dima-dencep: Uh? It was there [before](https://github.com/neoforged/FancyModLoader/blob/main/core/src/main/java/net/neoforged/fml/config/ConfigFileTypeHandler.java#L34), I just moved it down a bit.

core/src/main/java/net/neoforged/fml/config/ModConfig.java
Technici4n: @dima-dencep can you remove `dispatchConfigEvent` and `configHandler` from `ModContainer` as well?

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
Technici4n: Should we maybe disable it by default then for now? (at least in production)

core/src/main/java/net/neoforged/fml/config/ModConfig.java
dima-dencep: Done

dima-dencep: Everything looks good now, however the fact that the events are in a different subproject stresses me out a bit.... What is the purpose of this?

Technici4n: Until the eventbus changes made in 1.20.2, it was necessary for the events to be in the game layer to be processed by the eventbus transformer. This is technically not necessary anymore, but also not a problem in practice? In any case don't move them.

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
dima-dencep: How about a complete remove?

Technici4n: No - it would be nice to fix it eventually so please leave it in.

dima-dencep: Ok, what else do I need to do to get this pr merged? (I can't continue working on one mod without these changes)

embeddedt (approved):

Technici4n:
I'll do some testing later.

Technici4n (approved):

TheElectronWill:
Hello
You may be interested in knowing that https://github.com/TheElectronWill/night-config/pull/148 has been merged, and that https://github.com/TheElectronWill/night-config/pull/152 is in the works :)

PR #44, opened 2023-12-01
Remove `ModLoadingContext` and `FMLJavaModLoadingContext`

### LINKED ISSUE #39: Remove static `ModLoadingContext` access
The `ModLoadingContext` introduces a fair amount of global state through the field `private static final ThreadLocal<ModLoadingContext> context`, which is primarily used to resolve modids automatically in registration code. This usage should be removed, as global variables promote poor design and reduce readability of all code that reaches them.

There are two uses of this global state:
1. Automatic modid deduction (for registration of registry entries, configs, and client objects, and for IMC).
2. Provision of the language-specific loading context (i.e., `FMLJavaModLoadingContext`).

Case 1 is trivially resolvable. Modders should (and are very capable) of providing a modid to registration code at the time of registration.  At most it adds one method argument, and produces a much higher degree of readability and transparency, and the registered object's modid is not at the whim of a global state object.

Case 2 is resolvable by passing the language-specific loading context as a constructor argument to the target mod.

dima-dencep:
Now these parameters can be obtained through the constructor, `ModLoadingContext` and `FMLJavaModLoadingContext` are no longer needed

Resolves #39


Speiger:
Kinda against that.
Since i know mods that actually rely on that functionality to begin with.
Example:
https://github.com/Carbon-Config-Project/CarbonConfig/blob/1.20.2/src/main/java/carbonconfiglib/impl/internal/EventHandler.java#L83

Technici4n:
As you prefer

dima-dencep:
What to do with this pr after merge https://github.com/neoforged/FancyModLoader/pull/47?

PR #46, opened 2023-12-05
Reintroduce ModConfig#getHandler

pupnewfster:
Adds back the `ModConfig#getHandler` method that was removed in #40 so that mods can actually override the type handler.


Technici4n:
Not a fan of adding back a method that shouldn't exist. IMO, we should fix your use case properly by tweaking how configs are loaded.

marchermans:
> oved in #40 so that mods can actua

This was per review rejected.
The whole point of removing the handler is so that TOML is the default that should be used. It is practically the only one supported, and I see no reason to support other formats.

pupnewfster:
Superseded by #48

PR #47, opened 2023-12-05
Deprecate `FMLJavaModLoadingContext` for removal

dima-dencep:
As requested.

How do we handle this after `getModEventBus` got removed?
- https://github.com/neoforged/FancyModLoader/blob/main/languages/java/src/main/java/net/neoforged/fml/common/Mod.java#L81

What about updating other javadocs that have this method?


Technici4n:
> How do we handle this after getModEventBus got removed?

IMO we can remove the supplier from `Mod.EventBusSubscriber.Bus`, and instead pass the mod bus to `AutomaticEventSubscriber#inject`.

> What about updating other javadocs that have this method?

I would suggest linking `ModContainer#getEventBus` instead. But we are going to look into moving most events to the normal forge bus soon:TM:, so we can leave the old javadoc in place since it will have to change anyway.

PR #50, opened 2023-12-14
Implement dependency overrides

dima-dencep:
Wrote it in 30 minutes, didn't even test it (because I don't know how)

Inspired by [Fabric](https://fabricmc.net/wiki/tutorial:dependency_overrides)

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ModVersionOverrides.java
+++ b/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ModVersionOverrides.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) Forge Development LLC and contributors
+ * SPDX-License-Identifier: LGPL-2.1-only
+ */
+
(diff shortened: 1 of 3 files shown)


loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ModVersionOverrides.java
```diff
@@ -0,0 +1,110 @@
...
+        IConfigurable override = overrides.get(version.getModId());
+
+        if (override == null) {
+            return version;
+        }
+
+        return new OverridedModVersion(version, override);
+    }
+
+    public static class OverridedModVersion implements IModInfo.ModVersion {
```
Minecraftschurli, line 44: Overrided is wrong, it should be Overridden

Minecraftschurli (changes requested):

Technici4n:
After some discussion with other maintainers, we have decided that the idea of this feature is not desirable for FML. (On top of the specific implementation being suboptimal).

PR #51, opened 2023-12-14
Deprecate the `mandatory` dependency field and replace it with a `type` field

Matyrobbrt:
The mandatory field is verbatim replaced by `type="required"` and `type="optional"`.
Two more options are added:
- `incompatible`: prevents loading with the mod installed
- `discouraged`: prompts the user a warning when the mod is installed

A `reason` field is also added to be able to specify what leads to the incompatibility/conflict.

Depends on https://github.com/neoforged/NeoForgeSPI/pull/4

image

Something to consider later, when a rework of the loading code is done. The dependency graph should be able to distinguish between incompatible and normal dependencies for the sake of slightly improving the error message.


3TUSK:
> A `reason` field is also added to be able to specify what leads to the incompatibility/conflict.

Can you use translation key for this field?

Technici4n (approved):
🚢🇮🇹

Matyrobbrt:
> > A `reason` field is also added to be able to specify what leads to the incompatibility/conflict.
> 
> Can you use translation key for this field?

Unfortunately not @3TUSK, the error happens way too early for resources (and inherently mod lang files) to be loaded.

PR #56, opened 2023-12-17
Added monagstudios textures

mrlulu51:
I have created monagstudios texture as a joke.
Can you add it properly for April fools or by enabling it in the config ?


marchermans:
No sorry.

PR #57, opened 2023-12-18
Reduce resolution on mac

dima-dencep:


Technici4n:
Did you test this?

dima-dencep:
No, but that's how mixin works in rubidium-extra

I still don't have a mac, if anything...

PR #62, opened 2024-01-08
Remove naming related functionality from FML

shartte:
Now  that we remap to Mojmap runtime, I'd suggest we remove the functionality in FML that remaps intermediary<->Mojmap/MCP since it should not be functional anymore.

The `ObfuscationReflectionHelper` has been renamed to `ReflectionHelper` with the same interface. Naming-related functionality has been removed from it, and the existing class been marked as deprecated for removal.

This also removes an unused FML environment key `LOCATORCLASSLOADER` for which I could not find any uses.


embeddedt:
Hmm. Thoughts on keeping a `ReflectionHelper` around without remapping support? The API is nicer to use than pure reflection.

Matyrobbrt:
I think it makes sense to keep it, but renamed to `ReflectionHelper`.

PR #67, opened 2024-01-13
Remove hard casts to `ModFile` and co

Matyrobbrt:
Removes the hard casts to `ModFile`, to make `IModFile` extendable. In order to achieve this, a few more methods were added to SPI.


Su5eD (approved):
This looks great, we can finally get rid of all the needless hardcoding

marchermans (changes requested):
One question. Rest looks good

Technici4n:
My personal opinion is that making `IModFile` extendable is a bad idea. I don't like leaking details like the list of mixins for example.

PR #68, opened 2024-01-14
Fix wayland get window position

gravit0:
Fix bug https://github.com/neoforged/NeoForge/issues/393
This fix may hide other GLFW errors, but there is the same code nearby


earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/DisplayWindow.java
```diff
@@ -470,6 +470,7 @@ public void initWindow(@Nullable String mcVersion) {
         glfwGetFramebufferSize(window, x, y);
         this.fbWidth = x[0];
         this.fbHeight = y[0];
+        handleLastGLFWError((error, description) -> LOGGER.debug(String.format("Suppressing GLFW get window position error: [0x%X]%s", error, description)));
```
XFactHD, line 473: Shouldn't this rather be done directly after the offending method instead of after a second, mostly unrelated one in case the second one throws an error?

marchermans: Probably, @gravit0 Can you address this, so we can get this merged?

marchermans (approved):

PR #82, opened 2024-01-27
Add source info for JarInJar mod

TT432:
I tried to add the trace of the JarInJar mod to be able to print out the source of the JarInJar mod, which is useful for debugging in some complex environments, such as some environments with dozens of mods, because some non-standard packaging behavior or wrong java module definition causes conflicts or even crashes, in this case I need to find out the source of JarInJar to help identify the problem.


marchermans:
In general this is a way to solve it yes.
Not sure if I like the marking mechanism since you can get the same information from the uri of the jar in jar system. But a possible solution for sure.

TT432:
I've made changes to try printing the URI in the handleLocatedFiles, but my Git skills are not very good and my commits are a bit messy. Do I need to open a new pull request?

Technici4n:
We will squash, no worries.

marchermans (approved):

PR #84, opened 2024-01-30
[1.20.1] Publish using GH Actions

embeddedt:
To allow for future updates to be made while 1.20.1 is still supported.


.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
+name: Release
+
+on:
+  push:
+    branches: [ "main" ]
```
Technici4n, line 5: That's not gonna work for 1.20.1. I'd add it there.

PR #85, opened 2024-02-04
[1.20.1] Prevent mixins from crashing the game when there are missing mods

embeddedt:
The issue was fixed in newer FML versions by adding the new `[[mixin]]` syntax; this is a workaround for 1.20.1. I have copied the description from my original PR to MCF below.

----------------

In FML versions post-modularization, mixins are still loaded even if there is a missing dependency. Unfortunately, said mixins can attempt to load missing classes (causing a crash), or attempt to access a field that was transformed by an AT, which will not happen when there are loading errors, also causing a crash.

This PR solves the issue by not allowing mod resources to enter the classpath at all, preventing the mixin configs from being loaded which restores the 1.16 behavior. Now, the loading error screen will be properly shown instead of the game crashing with some random `ClassNotFoundException` or `IllegalAccessError`. This issue was particularly aggravating for support Discords as users report a "crash" when it's really just that they haven't installed a mod.

I have made the PR against 1.20.1, as that's the version I was using for testing. The test case I used in production is to install the mod Supplementaries, version 1.20-2.7.14, *without* installing its dependency, Moonlight Lib. Without this PR, the game crashes with `ClassMetadataNotFoundException`. If I replace the fmlloader JAR with one from this PR, the game reaches the loading error screen.

You'll notice that this change is disabled in dev. This is to prevent dev environments using `--mixin.config` from crashing if there is a dependency error due to the mixin file no longer being found. The issue is only really a problem for production environments, where users do not know to read their log and find the missing dependency error.

### DIFF
--- a/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModValidator.java
+++ b/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModValidator.java
@@ -18,6 +18,7 @@ import org.slf4j.Logger;
 import java.util.*;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class ModValidator {
     private static final Logger LOGGER = LogUtils.getLogger();
@@ -72,7 +73,7 @@ public class ModValidator {
     }
 
     public ITransformationService.Resource getModResources() {
-        Predicate<IModFile> shouldLoadResource;
+        Stream<ModFile> modFileStream = this.candidateMods.stream();
         if (FMLEnvironment.production) {
             // In production, only allow game libraries and/or mods in the loading mod list
             // This prevents custom mixins from loading if there is a dependency error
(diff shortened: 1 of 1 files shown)


loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModValidator.java
```diff
@@ -72,7 +72,22 @@ public ITransformationService.Resource getPluginResources() {
     }
 
     public ITransformationService.Resource getModResources() {
-        return new ITransformationService.Resource(IModuleLayerManager.Layer.GAME, this.candidateMods.stream().map(IModFile::getSecureJar).toList());
+        Predicate<IModFile> shouldLoadResource;
```
Technici4n, line 75: This is a cursed way of implementing this, I think it can be made simpler by building a `Stream<IModFile>` in each branch.

Technici4n (changes requested):

loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModValidator.java
```diff
@@ -72,7 +73,18 @@ public ITransformationService.Resource getPluginResources() {
...
+        Stream<ModFile> modFileStream = this.candidateMods.stream();
+        if (FMLEnvironment.production) {
+            // In production, only allow game libraries and/or mods in the loading mod list
+            // This prevents custom mixins from loading if there is a dependency error
+            // Usually, these mixins will break due to a missing class or AT, and that
+            // will prevent our error screen from ever becoming visible.
+            Set<IModFile> validMods = new HashSet<>();
+            validMods.addAll(this.loadingModList.getModFiles().stream().map(ModFileInfo::getFile).toList());
+            validMods.addAll(this.gameLibraries);
+            modFileStream = modFileStream.filter(validMods::contains);
```
Technici4n, line 85: Why not `validMods.stream()`?

embeddedt: Sure, same thing in the end.

Technici4n: FML code is already hard enough to read as is... :P

Technici4n (approved):

PR #86, opened 2024-02-04
[1.20.1] Parallelize background file scan

embeddedt:
Backport of https://github.com/neoforged/FancyModLoader/pull/32 to 1.20.1.

### DIFF
--- a/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/Scanner.java
+++ b/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/Scanner.java
@@ -34,7 +34,11 @@ public class Scanner {
         if (loaders != null) {
             loaders.forEach(loader -> {
                 LOGGER.debug(LogMarkers.SCAN, "Scanning {} with language loader {}", fileToScan.getFilePath(), loader.name());
-                loader.getFileVisitor().accept(result);
+                // Since we now scan files concurrently, but language loaders may not have been written with that in mind,
+                // synchronize on the language loader so it only scans one file at a time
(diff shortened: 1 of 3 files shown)


Technici4n:
What about the scan result future change? Also, is this a breaking change for language loaders?

embeddedt:
> What about the scan result future change?

I missed that, added the patch.

> Also, is this a breaking change for language loaders?

I'm very skeptical it would matter, but I added a `synchronized` block in the latest commit so any individual loader's file visitor will never be running concurrently on multiple files. This should be good enough.

loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/Scanner.java
```diff
@@ -34,7 +34,11 @@ public ModFileScanData scan() {
         if (loaders != null) {
             loaders.forEach(loader -> {
                 LOGGER.debug(LogMarkers.SCAN, "Scanning {} with language loader {}", fileToScan.getFilePath(), loader.name());
-                loader.getFileVisitor().accept(result);
+                // Since we now scan files concurrently, but language loaders may not have been written with that in mind,
+                // synchronize on the language loader so it only scans one file at a time
+                synchronized (loader) {
```
Technici4n, line 39: Doesn't this negate most of the benefits?

embeddedt: No, the hot code is `Scanner#fileVisitor`, which still runs concurrently.

Technici4n (approved):

PR #88, opened 2024-02-10
Add headers to crash report extenders

Matyrobbrt:
Neo already has a hook to add a crash report header, but the method contents are empty. This PR allows crash report extenders to add headers to crash reports.  

Connector can, for instance, use this to show a warning at the top of the report telling users to not report the crash to mod authors.

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/CrashReportCallables.java
+++ b/loader/src/main/java/net/neoforged/fml/CrashReportCallables.java
@@ -13,11 +13,13 @@ import java.util.Collections;
 import java.util.List;
 import java.util.function.BooleanSupplier;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 public class CrashReportCallables
 {
     private static final Logger LOGGER = LogUtils.getLogger();
     private static final List<ISystemReportExtender> crashCallables = Collections.synchronizedList(new ArrayList<>());
+    private static final List<Supplier<String>> HEADERS = Collections.synchronizedList(new ArrayList<>());
 
     /**
      * Register a custom {@link ISystemReportExtender}
@@ -90,8 +92,21 @@ public class CrashReportCallables
         });
     }
 
+    /**
+     * Registers a header to be added to the top of crash reports.
+     *
+     * @param header the header
+     */
+    public static void registerHeader(Supplier<String> header) {
+        HEADERS.add(header);
+    }
+
     public static List<ISystemReportExtender> allCrashCallables()
     {
         return List.copyOf(crashCallables);
+    }
+
+    public static Stream<String> getHeaders() {
+        return HEADERS.stream().map(Supplier::get);
     }
 }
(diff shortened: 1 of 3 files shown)


Su5eD (approved):

loader/src/main/java/net/neoforged/fml/ISystemReportExtender.java
```diff
@@ -5,14 +5,39 @@
...
+    @Override
+    default String get() {
+        return null;
+    }
+
+    /**
+     * {@return a header to add to the crash report}
+     */
+    @Nullable
+    default String getHeader() {
```
sciwhiz12, line 36: This is unused here. I assume there'll be a future NeoForge PR to implement usages of this?

Also, why is this nullable?

loader/src/main/java/net/neoforged/fml/ISystemReportExtender.java
```diff
@@ -5,14 +5,39 @@
...
+
 import java.util.function.Supplier;
 
-public interface ISystemReportExtender extends Supplier<String>
-{
-    String getLabel();
+@ApiStatus.OverrideOnly
+public interface ISystemReportExtender extends Supplier<String> {
+    /**
+     * {@return the label of the crash report value to add}
```
sciwhiz12, line 16: This isn't for the crash report but the _system_ report, as evidenced by the name of the class. 

While crash reports have a system report, a system report can be generated independently of a crash report -- see `PerfCommand` (`/perf`) where it saves a system report alongside the performance metrics report.

loader/src/main/java/net/neoforged/fml/ISystemReportExtender.java
```diff
@@ -5,14 +5,39 @@
...
+     */
+    @Nullable
```
sciwhiz12, line 18: Why is this and `get()` both nullable?

sciwhiz12 (changes requested):

loader/src/main/java/net/neoforged/fml/CrashReportCallables.java
```diff
@@ -90,8 +92,21 @@ public boolean isActive()
         });
     }
 
+    /**
+     * Registers a header to be added to the top of crash reports.
+     *
+     * @param header the header
+     */
+    public static void registerHeader(Supplier<String> header) {
```
Technici4n: custom functional interface?

loader/src/main/java/net/neoforged/fml/ICrashReportHeader.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Supplies a header to add to crash reports.
+ * @see CrashReportCallables#registerHeader(Supplier)
+ */
+@FunctionalInterface
+public interface ICrashReportHeader {
+    /**
+     * {@return the header to be displayed at the top of the crash report}
+     */
+    String getHeader();
```
Technici4n, line 19: should it be nullable in case there is no header to be added?

Technici4n (approved):

sciwhiz12 (approved):

PR #92, opened 2024-02-23
Look in the SERVICE layer for ImmediateWindowProvider implementations

FiniteReality:
This change was discussed in https://canary.discord.com/channels/313125603924639766/852298000042164244/1210370060737314818

There's a few ways to actually implement this, but I chose the most obvious. One behavioural change from this approach is that we no longer load ImmediateWindowProvider implementations from unnamed modules; if we're moving towards more use of modules, this seems like a sensible trade-off to me, and shouldn't pose a challenge for most mods as we build modules for them on the fly already.


loader/src/main/java/net/neoforged/fml/loading/ImmediateWindowHandler.java
```diff
@@ -32,7 +34,10 @@ public static void load(final String launchTarget, final String[] arguments) {
         } else {
             final var providername = FMLConfig.getConfigValue(FMLConfig.ConfigValue.EARLY_WINDOW_PROVIDER);
             LOGGER.info("Loading ImmediateWindowProvider {}", providername);
-            final var maybeProvider = ServiceLoader.load(ImmediateWindowProvider.class)
+            final var layer = Launcher.INSTANCE.findLayerManager()
+                    .flatMap(manager -> manager.getLayer(Layer.SERVICE))
+                    .orElse(ImmediateWindowHandler.class.getModule().getLayer()); // This will likely be the BOOT layer, but is a sensible fallback.
```
Technici4n, line 39: Is it a sensible fallback? I would be in favor of throwing an exception early on to detect misconfiguration / unexpected environments.

Technici4n (changes requested):
We also need to load jars that provide the service onto the SERVICE layer. Look at the two implementations of TransformerDiscoverySomething.

loader/src/main/java/net/neoforged/fml/loading/ImmediateWindowHandler.java
FiniteReality: I believe so - the original overload of `ServiceLoader.load` uses the current thread's context class loader, which should be the same one as set by the BOOT layer in `BootstrapLauncher`: <https://github.com/McModLauncher/bootstraplauncher/blob/main/src/main/java/cpw/mods/bootstraplauncher/BootstrapLauncher.java#L172>

Technici4n: I mean: is there any case in normal FML usage where the service layer would not be built by the team we reach that call? If not, we should throw. (Defensive programming and stuff)

FiniteReality: The call stack for a regular startup is something like this:
- BootstrapLauncher.main
- BootstrapLauncher.run
- BootstrapLauncherConsumer.accept
- Launcher.main
- Launcher.run
- TransformationServicesHandler.discoverServices
- ModDirTransformerDiscoverer.earlyInitialization
- ImmediateWindowHandler.load
And the SERVICE layer is built in `TransformationServicesHandler.discoverServices` just before calling `earlyInitialization`, which should make it impossible for it to happen. I personally don't expect `ImmediateWindowHandler.class.getModule().getLayer()` to ever be returned, but having it as a fallback should preserve the *current* behaviour if the order in which layers are built ever gets changed.

FiniteReality:
> We also need to load jars that provide the service onto the SERVICE layer. Look at the two implementations of TransformerDiscoverySomething.

Is this referring to the SERVICES set in `ModDirTransformerDiscoverer`? I was under the impression that anything providing a custom immediate window provider would likely also be providing an ITransformationService or similar - I'm definitely doing that myself

loader/src/main/java/net/neoforged/fml/loading/ImmediateWindowHandler.java
Technici4n: Once we make the change that `ImmediateWindowProvider`s are loaded from the SERVICE layer, this is a feature that we must preserve. As such, crashing if the SERVICE layer is not available is the most appropriate thing to do, so that we can immediately find a bug should the code be moved to a different place.

Technici4n:
> is there a reason why the two don't match

No idea! Looks like an oversight. We should deduplicate the set to avoid such mismatches in the future.

FiniteReality:
I've implemented the requested changes, and also extracted the service set into a constants class that gets shared between the two discoverers, which I've added the ImmediateWindowProvider type to. I think that at some point in the future we should have a breaking change to move it into the SPI project, to be more consistent with the other service types.

Technici4n:
Looks good, will need to test this somehow however 😅

marchermans (approved):

embeddedt (approved):
Simple and functional.

Technici4n:
Looks nice.

marchermans (approved):

marchermans:
> Looks nice.

I take this as an approval of the changes, and assumed that tech was unable to provide a proper approval due to external circumstances. Merging.

PR #93, opened 2024-02-25
Add support for locating mod sources with services from the classpath.

marchermans:
This allows a SERVICE-layer source set or jar to be loaded from the classpath in userdev if it is part of the MOD_CLASSES.


FiniteReality (approved):
This change seems fine as-is. Maybe we should consider extracting the parsing of MOD_CLASSES to a common location - I believe there's an exact duplicate of this logic in MinecraftLocator as well as one or two other places.

loader/src/main/java/net/neoforged/fml/loading/ClasspathTransformerDiscoverer.java
```diff
@@ -59,4 +70,38 @@ private void locateTransformers(String resource) throws IOException {
...
+    private void scanModClasses() {
+        final Map<String, List<Path>> modClassPaths = getModClasses();
+        modClassPaths.forEach((modid, paths) -> {
+            if (shouldLoadInServiceLayer(paths)) {
+                found.add(new NamedPath(modid, paths.toArray(Path[]::new)));
+            }
+        });
+    }
+
+    private Map<String, List<Path>> getModClasses() {
```
Technici4n, line 83: There is no need to duplicate this, I think we can make the method in `CommonLaunchHandler` static and call it directly.

loader/src/main/java/net/neoforged/fml/loading/ClasspathTransformerDiscoverer.java
```diff
@@ -59,4 +70,38 @@ private void locateTransformers(String resource) throws IOException {
             found.add(new NamedPath(path.toUri().toString(), path));
         }
     }
+
+    private void scanModClasses() {
+        final Map<String, List<Path>> modClassPaths = getModClasses();
+        modClassPaths.forEach((modid, paths) -> {
+            if (shouldLoadInServiceLayer(paths)) {
+                found.add(new NamedPath(modid, paths.toArray(Path[]::new)));
```
Technici4n, line 78: If a jar is loaded in the service layer like this, its modid should be blacklisted from being loaded by the normal mod locator. (similar to `ModDirTransformerDiscoverer#allExcluded()`)

Technici4n (changes requested):

loader/src/main/java/net/neoforged/fml/loading/ClasspathTransformerDiscoverer.java
marchermans: Right

loader/src/main/java/net/neoforged/fml/loading/TransformerDiscovererUtils.java
```diff
@@ -0,0 +1,23 @@
...
+import cpw.mods.jarhandling.JarContentsBuilder;
+import cpw.mods.jarhandling.JarMetadata;
+import cpw.mods.jarhandling.SecureJar;
+
+import java.nio.file.Path;
+
+public class TransformerDiscovererUtils {
+
+
+    public static boolean shouldLoadInServiceLayer(Path... path) {
```
Technici4n, line 17: I'd put it in the same class as `TransformerDiscovererConstants`, potentially renaming it.

Technici4n (approved):

loader/src/main/java/net/neoforged/fml/loading/ModDirTransformerDiscoverer.java
```diff
@@ -93,9 +90,6 @@ private static boolean shouldLoadInServiceLayer(Path path) {
         if (!path.toString().endsWith(".jar")) return false;
         if (LamdbaExceptionUtils.uncheck(() -> Files.size(path)) == 0) return false;
 
-        JarMetadata metadata = JarMetadata.from(new JarContentsBuilder().paths(path).build());
-        return metadata.providers().stream()
-            .map(SecureJar.Provider::serviceName)
-            .anyMatch(TransformerDiscovererConstants.SERVICES::contains);
+        return TransformerDiscovererConstants.shouldLoadInServiceLayer(path);
```
FiniteReality, line 93: Now that we're delegating to this method, is `shouldLoadInServiceLayer` a good name for this method? Would something like `isValidServiceLayerJar` be a better name?

loader/src/main/java/net/neoforged/fml/loading/moddiscovery/MinecraftLocator.java
```diff
@@ -47,7 +48,9 @@ public List<IModLocator.ModFileOrException> scanMods() {
                 .map(SecureJar::from)
                 .map(sj -> new ModFile(sj, this, ModFileParser::modsTomlParser))
                 .collect(Collectors.<IModFile>toList());
+        var otherModsExcluded = ClasspathTransformerDiscoverer.allExcluded();
         var othermods = baseMC.otherModPaths().stream()
+                .filter(p -> p.stream().noneMatch(otherModsExcluded::contains)) //We can no load MOD_CLASSES from the classpath if they are loaded on the SERVICE layer.
```
FiniteReality, line 53: Is `no load` a typo here?

loader/src/main/java/net/neoforged/fml/loading/ModDirTransformerDiscoverer.java
Technici4n: I think it's fine like this. Find a synonymous name doesn't make the method clearer.

Technici4n (approved):

PR #94, opened 2024-03-01
Switch to Fabric Mixin

Matyrobbrt:
This PR switches FML to use Fabric Mixin instead of Sponge's.

## Background
Mixin's last release was in December 2021, and until 2 months ago it saw no public commits. In late January, a 0.8.6 snapshot was pushed, but it has several issues and missing features (an injection point was broken, and the new mixin source feature isn't actually used), and we have seen no progress on fixing them.  
Mixin has never shown a desire to bring on more than its current one maintainer (who has little time available), and PRs don't end up being merged for the same time reason. Moreover, our communication with the Mixin maintainer is horrendous, not only because there are few channels of communication (they use a Discord IRC bridge that only shows them 2 channels from our Discord server), but also because we have been pinging them constantly with no reply in weeks, and it has been that way for years.  

Additionally, we have also promised in our [2023 end-of-year blog post](https://neoforged.net/news/2023-retrospection/) to look into alternatives. It's been 2 months, it's about time.

## The Fabric fork
Fabric has [their own fork](https://github.com/fabricmc/mixin) of Mixin, which has several features that "stock" mixin doesn't provide (although some are provided in the new snapshot, but they either cause other issues, or are simply not fully implemented):
- injections in interface methods
- automatic mixin method prefixing with the mod ID
-  fewer restrictions when it comes to constructor injection (stock mixin allows only RETURN targeting)
- array creations in field initialisers
- CIs are only created when needed, improving performance

Using the fork comes with no penalty for us, as we can proxy their maven and forever store their mixin artifacts, without allowing them to be changed.  
Assuming Mixin will eventually have all of the features Fabric's mixin has, we can switch back to Sponge Mixin without causing any issues.

## Testing
You can test that interface injections and mixin prefixing works by writing a mixin that mixes into a vanilla interface, and that throws an exception. The mod ID should show up in the stacktrace.

## Testing previous behaviour
To test that Fabric's version of Mixin does not break intended Sponge Mixin behaviour, you can run with [Mixin Booster](https://curseforge.com/minecraft/mc-mods/mixinbooster) in 1.20.1 packs

See https://github.com/Sinytra/Connector/discussions/383 too for more information.


TelepathicGrunt:
I oppose!..

…to not merging this. 

Looks good to me

TheCurle:
I feel like Fabric Mixin is the better option over maintaining our own fork:

1) We're already limited on available manpower.

2) Fabric Mixin is extensively tested to provide what people want.

3) Fabric Mixin is in a pseudostable state, no further features or changes are expected to be made to it.

4) The intricacies and potential pain points of Fabric Mixin are well documented.

I feel like all of these are pure advantages and together form a good reason to use that over a custom fork, especially since getting a personal fork working will take yet more time to implement.

marchermans:
> I feel like Fabric Mixin is the better option over maintaining our own fork:
> 
>     1. We're already limited on available manpower.
> 
>     2. Fabric Mixin is extensively tested to provide what people want.
> 
>     3. Fabric Mixin is in a pseudostable state, no further features or changes are expected to be made to it.
> 
>     4. The intricacies and potential pain points of Fabric Mixin are well documented.
> 
> 
> I feel like all of these are pure advantages and together form a good reason to use that over a custom fork, especially since getting a personal fork working will take yet more time to implement.

That seems well documented and thought through. Perfect. I will approve the PR then

TheCurle (approved):

loader/build.gradle
```diff
@@ -27,7 +27,7 @@ dependencies {
     implementation("net.neoforged.accesstransformers:at-modlauncher:${accesstransformers_version}") {
         exclude module: 'modlauncher'
     }
-    implementation("org.spongepowered:mixin:${mixin_version}")
+    implementation("net.fabricmc:sponge-mixin:${mixin_version}")
```
marchermans, line 30: I assume this is available on maven central?

Matyrobbrt: It is not, I have set up a proxy for that specific group on our maven.

marchermans (approved):

loader/build.gradle
marchermans: Even better!

PR #99, opened 2024-03-20
Add `registerConfig` method to `ModContainer`

dhyces:
Adds a method to `ModContainer` to create and add a mod config from a given type and spec. This is added to simplify the current verbose `addConfig` method. I understand that the config system is currently being looked at for a future refactor and as such recognize that the lifetime of this method may be short.

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/ModContainer.java
+++ b/loader/src/main/java/net/neoforged/fml/ModContainer.java
@@ -160,6 +160,10 @@ public abstract class ModContainer
         addConfig(new ModConfig(type, configSpec, this));
     }
 
+    public void registerConfig(ModConfig.Type type, IConfigSpec<?> configSpec, String fileName) {
+        addConfig(new ModConfig(type, configSpec, this, fileName));
+    }
+
     /**
      * Does this mod match the supplied mod?
      *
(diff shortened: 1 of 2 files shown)


Shadows-of-Fire (approved):

ChampionAsh5357:
I would also add the method to change the filename of the config and redirect the `ModLoadingContext` methods to use these new methods within the container.

dhyces:
> I would also add the method to change the filename of the config and redirect the `ModLoadingContext` methods to use these new methods within the container.

Yeah, I don't see why not. Aside from that, should anything be done with the validation currently in the `ModLoadingContext` methods? Could move it into `addConfig`, unless there is a case to register an empty mod config, or can change it to only log a warning without early return.

ChampionAsh5357:
> Yeah, I don't see why not. Aside from that, should anything be done with the validation currently in the `ModLoadingContext` methods? Could move it into `addConfig`, unless there is a case to register an empty mod config, or can change it to only log a warning without early return.

Since `#addConfig` didn't have the validation in the first place, I would leave it alone. However, they can be moved to the two methods you created.

Technici4n (changes requested):
Could you please move the `isEmpty` check to the new methods as well, and deprecate the old methods for removal?

Technici4n (approved):

PR #102, opened 2024-03-21
[1.20.1] Backport some fixes from main

embeddedt:
This PR makes two changes:

* Catch `NoSuchFileException` as well as `FileNotFoundException` in `AbstractJarFileDependencyLocator`. This allows running with the latest SJH without seeing a few ugly errors on startup (they are harmless, but we may as well hide them).
* Backport #95 to 1.20.1, as the performance issue also exists on that version.

### DIFF
--- a/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/AbstractJarFileDependencyLocator.java
+++ b/loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/AbstractJarFileDependencyLocator.java
@@ -14,6 +14,7 @@ import org.slf4j.Logger;
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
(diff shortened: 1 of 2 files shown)


loader/src/main/java/net/minecraftforge/fml/loading/moddiscovery/AbstractJarFileDependencyLocator.java
```diff
@@ -25,7 +26,7 @@ protected Optional<InputStream> loadResourceFromModFile(final IModFile modFile,
         try {
             return Optional.of(Files.newInputStream(modFile.findResource(path.toString())));
         }
-        catch (final FileNotFoundException e) {
+        catch (final FileNotFoundException | NoSuchFileException e) {
             LOGGER.debug("Failed to load resource {} from {}, it does not contain dependency information.", path, modFile.getFileName());
```
Technici4n, line 30: don't we want to `trace` this?

PR #105, opened 2024-03-27
Remove FMLModType.LANGPROVIDER (you can use LIBRARY instead)

### LINKED ISSUE #71: Deprecate/remove LANGPROVIDER jar type
It is equivalent to the `LIBRARY` type at the moment - since we load the langproviders via ServiceLoader, we load them from LIBRARY jars too.

Technici4n:
There was in fact no reason to have this separate mod type - it didn't do anything special.

Closes #71.


thedarkcolour (approved):
Good change, keeping LANGPROVIDER and LIBRARY code separate is tedious.

marchermans (approved):

PR #110, opened 2024-04-03
Vastly simplify mod initialization

Technici4n:
Redo mod loading to remove a lot of unnecessary indirection, and finally make the code understandable.

All the steps are now handled by straightforward methods in `ModLoader` such as `dispatchParallelEvent` or `runInitTask`. Dispatch happens in a single NeoForge class.

See https://github.com/neoforged/NeoForge/pull/774.


Elec332:
There is a usecase (that would need s simple hook) that will no longer be possible after this PR, namely getting a callback after the modlist (`net.neoforged.fml.ModList`) has been set (AKA after all mods have been gathered).

Previously this was (perhaps unintentionally) possible with the following:
```
public class NeoStatesProvider implements IModStateProvider {

    @Override
    public List<IModLoadingState> getAllStates() {
        return List.of(ModLoadingState.withInline("POPULATE_MODLIST", "", ModLoadingPhase.GATHER, ml -> LoaderInitializer.INSTANCE.finalizeLoading()));
    }

}
```
(Until recently this had the downside that an `IModStateProvider` couldn't be loaded from a `LANGPROVIDER` so it had to be loaded from a JIJ'ed mod, but I just noticed this is no longer the case in NF)

The usecase for me specifically is a sub-loader that supports Neo, Forge & Fabric (and Quilt).
Like any modloader it builds a (mirrored & abstracted) list of containers before doing any initialization stuff.
This method allowed me to finalize the loader before going into the mod lifecycle (constructing & events).

I do have a `net.neoforged.neoforgespi.language.IModLanguageProvider` that does the containers for NF, which is relevant when the modlist is getting built. (https://github.com/Elecs-Mods/ElecLoader/blob/master/neo/src/main/java/nl/elec332/minecraft/loader/impl/neolang/NeoModContainer.java)

But when I have build the containers for NF I would now have no more options to hook into FML to make sure I can also built my modlist before the mods get constructed.

Technici4n:
@Elec332 thanks for the reply.

> This method allowed me to finalize the loader before going into the mod lifecycle (constructing & events).

I don't have the impression that it must be done like that. Couldn't you build the mod list on-demand? For example the first time it is queried. You could also do the loader finalization the first time an ElecModContainer is built.

Elec332:
While a part of it could be done lazily like that (and I considered it before I found the custom loader state way worked, but I really disliked it), it is really not a nice way of doing things.

First of all saying "the first of an unknown number of mods is being loaded, so we're building the modlist I guess, let's do part of the finalization" is not a great way of doing things. 
Secondly, the loader does some sanity checks against the "real" and mapped list and also makes sure that all custom ModContainers have been loaded correctly. 
Lastly, finalizing the modlist lazily is in this case an ugly way to do it. When the first opportunity arises that the loader can be queried by a mod it should be fully finalized & immutable, just like all other modloaders.
(Also, because loading events are parallel it can block multiple mods while the loader is still lazily initializing) 

It is currently possible with all (big) loaders; Fabric (and Quilt) makes their containers so early that a `PreLaunchEntrypoint` can lock in the modlist (and to make sure loading behavior is consistent between all loaders it my loader mimics the Neo/Forge way of loading mods later on, so it ends up not needing it that early) and for Forge I can use the above described method.

Neo would in this case be the one loader where it would be impossible to fully finilize the loader before any mods get loaded, which considering the amount of hooks available for custom 'stuff' in FML would seem a bit strange to me.

A nice place for something like this could be a (default) method in `IModLanguageProvider`.
It seems like there used to be ways to hook into the lifecycle, which are no longer in use:
```
default <R extends ILifecycleEvent<R>> void consumeLifecycleEvent(Supplier<R> consumeEvent) {
}
```

Since I guess lifecyle events were replaced by the `IModStateProvider` stuff (and a bit more of course) it would seem like there have been ways to hook into the lifecycle for a long time until now. Perhaps a reconsideration of a similar hook for the lifecycle, or for this case, a default method in `IModLanguageProvider` like
```
    default void afterAllModsGathered() {
    }
```
which would be called right before the new `constructMods` method could be considered.

Technici4n:
@Elec332 I added `IModListReadyCallback` which should be a suitable replacement for your use case.

Elec332:
Looks good, thanks!

Shadows-of-Fire:
Well I don't entirely know what structural changes are going on here, and it's non-obvious from just browsing the changeset, but I don't see anything egregious

loader/src/main/java/net/neoforged/fml/ModLoader.java
```diff
@@ -191,73 +190,108 @@ public void gatherAndInitializeMods(final ModWorkManager.DrivenExecutor syncExec
...
-                    .map(ModLoadingException.class::cast)
-                    .collect(Collectors.toList());
-            LOGGER.fatal(LOADING, "Failed to complete lifecycle event {}, {} errors found", state.name(), modLoadingExceptions.size());
-            statusConsumer.ifPresent(c -> c.accept("ERROR DURING MOD LOADING"));
-            throw new LoadingFailedException(modLoadingExceptions);
+    private void waitForFuture(String name, Runnable ticker, CompletableFuture<?> future) {
+        while (true) {
+            ticker.run();
+            try {
+                future.get(50, TimeUnit.MILLISECONDS);
```
Shadows-of-Fire, line 273: Is it guaranteed that all futures being resolved in this method are expected to take less than 50ms?  Seems like a fairly short timeout, but I don't know what's going through here.

loader/src/main/java/net/neoforged/fml/loading/progress/StartupNotificationManager.java
```diff
@@ -31,22 +31,22 @@ public static List<ProgressMeter> getCurrentProgress() {
...
+            progressMeters.addLast(pm);
         }
         return pm;
     }
 
     public static void popBar(final ProgressMeter progressMeter) {
         synchronized (progressMeters) {
-            progressMeters.remove(progressMeter);
+            progressMeters.removeLastOccurrence(progressMeter);
         }
```
Shadows-of-Fire, line 50: Why were these reordered?  Were they incorrect previously?

loader/src/main/java/net/neoforged/fml/ModLoader.java
embeddedt: The idea seems to be to use a short timeout so that the loop will go around and execute the ticker again every 50ms. However, I'm not sure throwing a `TimeoutException` every 50ms is ideal. Constructing exceptions is not fast.

Technici4n: The ticker is what does the loading screen re-draw. This controls the FPS of the loading screen basically. Exceptions are not cheap but this should be fine.

loader/src/main/java/net/neoforged/fml/loading/progress/StartupNotificationManager.java
Technici4n: Yes. This makes new progress bars render below old ones. (Previously a new one would render higher than an old one)

embeddedt: > This makes new progress bars render below old ones. (Previously a new one would render higher than an old one)

Isn't that what we want? For instance, if you had a main progress bar for each registry, and then another one showing progress through the mod list, I would argue the mod one should be shown more prominently since it moves faster and has more useful information.

Technici4n: As discussed on Discord we want the outer step to be above the inner step.

loader/src/main/java/net/neoforged/fml/ModLoader.java
Technici4n: I know - the goal is just to have _something_. The background waiter does the same. I'm happy to change it in a future PR.

PR #111, opened 2024-04-03
Update foojay and refactor `EventBusSubscriber` as an outer class

### LINKED ISSUE #101: Move `EventBusSubscriber` into outer class
`EventBusSubscriber` is currently an inner class of `Mod` for no obvious reason. This creates a double inner class and makes `EventBusSubscriber` less convenient to use. Additionally, according to Orion, both `Mod` and `EventBusSubscriber` are part of the Java language provider specification, yet these classes are in the `common` package. I propose that `Mod` be moved to the `javafmlmod` package and `EventBusSubscriber` be moved to an outer class in the same package.

Spinoscythe:
Update foojay and refactor `EventBusSubscriber` as an outer class
Fixes #101


Technici4n (changes requested):
This needs more thought than just blindly applying an issue. Chances are that other language providers are also using these classes.

Spinoscythe:
> FML needs more global vision than that
- @Technici4n via discord in #neoforge-github

PR #116, opened 2024-04-17
Possibility to modify NeoForge group

TonimatasDEV:
I created this PR to add an argument ("--fml.neoForgeGroup") to modify the NeoForge artifact group. You don't need to add this argument to win_args and unix_args because when the argument is null automatically sets "net.neoforged".

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/loading/FMLServiceProvider.java
+++ b/loader/src/main/java/net/neoforged/fml/loading/FMLServiceProvider.java
@@ -36,6 +36,7 @@ public class FMLServiceProvider implements ITransformationService
     private ArgumentAcceptingOptionSpec<String> mcOption;
     private ArgumentAcceptingOptionSpec<String> mcpOption;
     private ArgumentAcceptingOptionSpec<String> mappingsOption;
+    private ArgumentAcceptingOptionSpec<String> groupOption;
     private List<String> modsArgumentList;
     private List<String> modListsArgumentList;
     private List<String> mavenRootsArgumentList;
@@ -45,6 +46,7 @@ public class FMLServiceProvider implements ITransformationService
     private String targetMcVersion;
     private String targetMcpVersion;
     private String targetMcpMappings;
+    private String neoForgeGroup;
     private Map<String, Object> arguments;
 
     public FMLServiceProvider()
@@ -76,6 +78,7 @@ public class FMLServiceProvider implements ITransformationService
         arguments.put("mcVersion", targetMcVersion);
         arguments.put("neoFormVersion", targetMcpVersion);
         arguments.put("mcpMappings", targetMcpMappings);
+        arguments.put("neoForgeGroup", neoForgeGroup);
         LOGGER.debug(CORE, "Preparing launch handler");
         FMLLoader.setupLaunchHandler(environment, arguments);
         FMLEnvironment.setupInteropEnvironment(environment);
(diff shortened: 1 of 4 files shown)


Technici4n:
May I ask how you intend to use this?

C0D3-M4513R:
The intention is to allow other people to release and launch custom versions of neoforge.
Releasing neoforge under a foreign group, just because we have to is not the right approach imo.
Ketting e.g. uses a similar argument on forge’s fml launcher.

C0D3-M4513R:
Though it honestly is not too much of an issue, should you not accept this pr.
That will just mean, that we have to release neo’s fml ourselves (similarly to forge).

dima-dencep:
Someday people will realize that merging bukkit binpatches into neoForge is a terrible idea, and start using mixins

C0D3-M4513R:
mixin based hybrids, like archlight and mohist have been reported to not be as stable, as source patched hybrids like magma or ketting.

dima-dencep:
If you at least knew how mixins work, you wouldn't have taken up binarypatch, and the main problem with archlight is that the mixins there are hastily written, but if you have at least a little bit of brains, you won't have any trouble fixing that

dima-dencep:
Archlight is just a mod with its own launcher

C0D3-M4513R:
Making changes to multiple, different parts in a function, whilst relying on function local variables seems impossible with mixins.

dima-dencep:
Obviously, if this was required for bukkit, archlight would not exist

dima-dencep:
I don't want to discuss hybrid cores here anymore, you better stay on lexforge and die with it, sorry

JustRed23:
> I would not like to merge this since the real way is to make a custom launch target

This was a part of the original lexforge code though, this is basically just a reimplementation

C0D3-M4513R:
> I would not like to merge this since the real way is to make a custom launch target

Also, I think that a custom launch target would not allow us to change where the server binpatches and universal jar get loaded from.

JustRed23:
> I don't want to discuss hybrid cores here anymore, you better stay on lexforge and die with it, sorry

Not to be rude or anything, but you started the discussion. Sure mixins are pretty nice, but not well documented (not talking about the fabric mixin docs). Not everyone with Java knowledge knows how bytecode works, so directly implementing it into the project makes it easier for contributors who do not know how mixins work to help debugging.

C0D3-M4513R:
Though tbf, if ketting only used mixins, it would be a lot easier to port.
I do see a future re-implementation of ketting using only mixins.

JustRed23:
> May I ask how you intend to use this?

The PR will be primarily used in our project to relocate the patched jars to another forlder, so instead of libraries/net/neoforged/.. it could be something like libraries/com/example/...

So basically undoing a hardcoded value

JustRed23:
> > > I would not like to merge this since the real way is to make a custom launch target
> > 
> > 
> > Also, I think that a custom launch target would not allow us to change where the server binpatches and universal jar get loaded from. Please correct me, if I'm wrong tho.
> 
> For the client: https://github.com/neoforged/FancyModLoader/blob/main/loader/src/main/java/net/neoforged/fml/loading/targets/ForgeClientLaunchHandler.java#L22
> 
> For the server: https://github.com/neoforged/FancyModLoader/blob/main/loader/src/main/java/net/neoforged/fml/loading/targets/ForgeServerLaunchHandler.java#L22

yea these are literally the locations that are hardcoded to net.neoforged, this is why the PR exists, as it's a FML thing

C0D3-M4513R:
> > > I would not like to merge this since the real way is to make a custom launch target
> > 
> > 
> > Also, I think that a custom launch target would not allow us to change where the server binpatches and universal jar get loaded from. Please correct me, if I'm wrong tho.
> 
> For the client: https://github.com/neoforged/FancyModLoader/blob/main/loader/src/main/java/net/neoforged/fml/loading/targets/ForgeClientLaunchHandler.java#L22
> 
> For the server: https://github.com/neoforged/FancyModLoader/blob/main/loader/src/main/java/net/neoforged/fml/loading/targets/ForgeServerLaunchHandler.java#L22

Ah, I see now. Thanks.

PR #112, opened 2024-04-08
Locator rewrite

Matyrobbrt:
The overall goal of this PR is to make the code structure in the FML mod discovery process easier to follow and to support a better end-to-end error reporting experience for mod loading issues of all kinds.

To this end, the PR rewrites the mod discovery system with the following changes:

- Remove superfluous API surface
- Add test coverage to the mod discovervy process by trying to simulate our common installation environments (Production, NeoForge Developoment, Mod Development)
- Unify the reporting of mod loading errors and warnings on a single type `ModLoadingIssue`, and provide a single exception that can report one or more such issues up the call chain (`ModLoadingException`)
- Allow locators to produce output at different levels of the discovery pipeline:
  - Unprocessed filesystem locations (i.e. from files in the `mods` folder)
  - Virtual Jar Files combined from various file-system locations and filtered (i.e. userdev locator that splits the combined Minecraft+Neoforge Jar into multiple separate virtual Jar files)
  - Fully built mod files (such as the Minecraft mod-file for which the locator provides the mod metadata on-the-fly)
- Allow locators to also supply mod loading issues (warnings and errors) on a per-file basis
- Introduce a priority system for locators and readers, allowing user supplied implementations to go before/after the built-in ones
- Untangle the dependency between locators and launch targets by having launch targets supply additional locators that are specific to the environment they are concerned with (i.e. the production launch targets supply locators for the production client/server)
- Untagle the dependency on the early phase ModLauncher discovery by determining whether a file has already been loaded not by reaching into previous locators statically, but rather by inspecting the ModLauncher ModuleLayers that have already been built when FML starts mod discovery.

PR #121, opened 2024-04-27
Bunch of small changes

xtrm-en:
This is mainly fixing some bottom-fruit warnings and adding paragraph tags to documentation for better readability (in IntellIJ at least).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,21 +76,26 @@ subprojects { subProject ->
     subProject.version = rootProject.version
 
     jar.doFirst {
+        var subName = subProject.name
+        var subVersion = subProject.version.toString()
+        var subVersionData = subVersion.split("\\.")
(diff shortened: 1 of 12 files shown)


xtrm-en:
There's indeed some small conflicts, would it be okay for me to rebase this PR onto #112's branch then?

Technici4n (changes requested):
We generally do not accept low hanging fruits PRs such as this one. Usually they bring very little value to the project, and accepting them would lead to more people attempting to make them.

xtrm-en:
> We generally do not accept low hanging fruits PRs such as this one. Usually they bring very little value to the project, and accepting them would lead to more people attempting to make them.

Understandable as it can be a lot to manage. Maybe adding this to a CONTRIBUTORS file would be interesting?

PR #123, opened 2024-05-02
Implement dependency-based (partial) ordering for parallel initialization tasks

Technici4n:
This PR allows mods to be constructed etc. _after_ their dependencies are. (Assuming `ordering="AFTER"` of course; otherwise the opposite happens). To do so, this PR does the following:
- Change mod ordering so that each mod inside a file is ordered independently from other mods (previously only files as a whole were ordered). This allows us to test this ordering in NeoForge, and it generally makes more sense given that dependencies are per-mod rather than per-file.
- Cache dependency information in adjacency list format in the `LoadingModList`.
- Do a bit of `CompletableFuture` manipulation to ensure that parallel tasks respect dependency "links". Mods with ordering constraints will respect the constraints, whereas "unrelated" mods will run in parallel.

Fixes https://github.com/neoforged/NeoForge/issues/540.

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/ModLoader.java
+++ b/loader/src/main/java/net/neoforged/fml/ModLoader.java
@@ -225,7 +225,7 @@ public final class ModLoader {
                                     if (exception != null) {
                                         // If there was any exception, short circuit.
                                         // The exception will already be handled by `waitForFuture` since it comes from another mod.
-                                        LOGGER.error("Skipping {} future for mod {} because a dependency threw an exception.", name, modContainer.getModId());
+                                        LOGGER.debug("Skipping {} task for mod {} because a dependency threw an exception.", name, modContainer.getModId());
                                         progress.increment();
                                         // Throw a marker exception to make sure that dependencies of *this* task don't get executed.
                                         throw new DependentFutureFailedException();
--- a/loader/src/main/java/net/neoforged/fml/ModLoader.java
+++ b/loader/src/main/java/net/neoforged/fml/ModLoader.java
@@ -207,7 +207,7 @@ public final class ModLoader {
             Map<IModInfo, CompletableFuture<Void>> modFutures = new IdentityHashMap<>();
             var futureList = modList.getSortedMods().stream()
                     .map(modContainer -> {
-                        // Build future for all dependencies first
+                        // Collect futures for all dependencies first
                         var deps = LoadingModList.get().getDependencies(modContainer.getModInfo());
                         @SuppressWarnings("unchecked")
                         CompletableFuture<Void>[] depFutures = new CompletableFuture[deps.size()];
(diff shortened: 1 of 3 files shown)


loader/src/main/java/net/neoforged/fml/ModLoader.java
```diff
@@ -190,22 +191,58 @@ public static void waitForTask(String name, Runnable periodicTask, CompletableFu
...
+                        CompletableFuture<Void>[] depFutures = new CompletableFuture[deps.size()];
+                        for (int i = 0; i < deps.size(); ++i) {
+                            depFutures[i] = modFutures.get(deps.get(i));
+                            if (depFutures[i] == null) {
+                                throw new IllegalStateException("Dependency future for mod %s which is a dependency of %s not found!".formatted(
+                                        deps.get(i).getModId(), modContainer.getModId()));
+                            }
+                        }
+
+                        var combinedFuture = depFutures.length == 0 ? CompletableFuture.completedFuture(null) : CompletableFuture.allOf(depFutures);
```
embeddedt, line 222: You can just use `CompletableFuture.allOf` unconditionally; it [handles](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#allOf-java.util.concurrent.CompletableFuture...-) zero-length arrays.

Technici4n: Done :)

embeddedt (approved):

embeddedt (approved):

loader/src/main/java/net/neoforged/fml/ModLoader.java
```diff
@@ -190,22 +191,57 @@ public static void waitForTask(String name, Runnable periodicTask, CompletableFu
...
+                                })
+                                .toArray(CompletableFuture[]::new);
+
+                        // Build the future for this container
+                        var future = CompletableFuture.allOf(depFutures)
+                                .<Void>handleAsync((void_, exception) -> {
+                                    if (exception != null) {
+                                        // If there was any exception, short circuit.
+                                        // The exception will already be handled by `waitForFuture` since it comes from another mod.
+                                        LOGGER.error("Skipping {} future for mod {} because a dependency threw an exception.", name, modContainer.getModId());
```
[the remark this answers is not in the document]
Technici4n: So, there are two alternatives here (behavior wise):
- We just catch the error and let dependent tasks execute normally. This can lead to errors in the dependent task, and seems quite pointless anyway.
- We let the error propagate via `.thenRunAsync`. This will lead to the same "issue" being displayed multiple times.

So the best course of action for us is to ignore the error. But I wanted to make it clear that we are explicitly bypassing the mod, rather than just silently swallowing the error.

PR #130, opened 2024-05-04
Consider all completed module layers when looking up services

### LINKED ISSUE #129: IModLanguageProviders not detected by LanguageProviderLoader
After the FMLLocator rewrite, IModLanguageProviders are no longer being detected by the LanguageProviderLoader. Its service detection code looks something like this:
```java
// LanguageProviderLoader.<init>
ServiceLoaderUtil.loadServices(launchContext, IModLanguageProvider.class)
// ServiceLoaderUtil.loadServices
// ...
var serviceLoader = context.createServiceLoader(serviceClass);
// ...
// LaunchContext.createServiceLoader
public <T> ServiceLoader<T> createServiceLoader(Class<T> serviceClass) {
    var moduleLayer = moduleLayerManager.getLayer(IModuleLayerManager.Layer.SERVICE).orElseThrow();
    return ServiceLoader.load(moduleLayer, serviceClass);
}
```
Before, the code was slightly different, scanning for services on the `PLUGIN` module layer rather than the `SERVICE` layer:
```java
// LanguageLoadingProvider
LanguageLoadingProvider() {
    var sl = Launcher.INSTANCE.environment().findModuleLayerManager().flatMap(lm->lm.getLayer(IModuleLayerManager.Layer.PLUGIN)).orElseThrow();
    serviceLoader = ServiceLoader.load(sl, IModLanguageProvider.class);
    loadLanguageProviders();
}
```
Since language providers are packaged as `FMLModType.LIBRARY` mod files, their service are only accessible at the `PLUGIN` layer. It was mentioned on the Discord that this was a bug, since language loaders that worked prior to the update are no longer being detected by NeoForge. 

If you are looking for a language provider to test with, I found this issue with Kotlin for Forge.

shartte:
Fixes #129

### DIFF
--- a/loader/src/test/java/net/neoforged/fml/test/RuntimeCompiler.java
+++ b/loader/src/test/java/net/neoforged/fml/test/RuntimeCompiler.java
@@ -10,10 +10,13 @@ import java.io.IOException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import javax.tools.DiagnosticCollector;
 import javax.tools.JavaCompiler;
 import javax.tools.JavaFileObject;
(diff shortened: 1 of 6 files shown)


loader/src/test/java/net/neoforged/fml/test/RuntimeCompiler.java
```diff
@@ -25,18 +28,28 @@
...
     private static final Logger LOGGER = LoggerFactory.getLogger(RuntimeCompiler.class);
     private static final JavaCompiler COMPILER = ToolProvider.getSystemJavaCompiler();
 
     private final DiagnosticCollector<JavaFileObject> diagnostics;
     private final StandardJavaFileManager manager;
+    private final Path rootPath;
+    private FileSystem openedFileSystem;
+
+    public static RuntimeCompiler create(Path targetFile) throws IOException {
+        var fs = FileSystems.newFileSystem(URI.create("jar:" + targetFile.toUri()), Map.of("create", true));
```
Matyrobbrt, line 41: Can you not pass the path directly instead of an uri?

Matyrobbrt (changes requested):

Matyrobbrt (approved):

PR #135, opened 2024-05-06
Allow IModLanguageLoaders to specify their version

thedarkcolour:
This PR addresses two problems with the current implementation.

Currently, the version of a language loader is found using the following code in LanguageProviderLoader:
```java
Optional<String> version = JarVersionLookupHandler.getVersion(lp.getClass());
String impl = version.orElse(Files.isDirectory(lpPath) ? FMLLoader.versionInfo().fmlVersion().split("\\.")[0] : null);
```
In userdev, JarVersionLookupHandler.getVersion returns empty, because the module is a directory with a name like `main` or `test`, which does not include a version number (the reasoning behind this is in JarMetadata.computeNameAndVersion, which works similarly to Java's [ModuleFinder](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/module/ModuleFinder.html#of-java.nio.file.Path...-l)). 

Prior to the language loader rewrite, the version was actually determined by the `Implementation-Version` specified in the MANIFEST.MF file, which could be loaded in userdev. Although non-intuitive and a bit obscure, this did not cause issues. Now, specifying one in userdev is impossible.

Neo currently hardcodes a workaround that works for its built-in language loaders, but not for any modded ones that don't adhere to FML's version numbers (Kotlin for Forge is on 5.x in 1.20.6). As of now, `FMLLoader.versionInfo().fmlVersion().split("\\.")[0]` always returns 3, which for Kotlin for Forge means Minecraft 1.18-1.19.2 (incorrect version).

To solve both of these issues, this PR removes the hardcoded version detection, allowing IModLanguageLoader to specify their version just as they'd specify their name. The old (from the language loader rewrite) code has been moved to a default method in the IModLanguageLoader interface, so there are no breaking changes.


loader/src/main/java/net/neoforged/neoforgespi/language/IModLanguageLoader.java
```diff
@@ -31,10 +37,23 @@
  */
 public interface IModLanguageLoader {
     /**
-     * {@return the name of this loader, used to decide what loader should load a mod}
+     * @return the name of this loader, used to decide what loader should load a mod
```
Matyrobbrt, line 40: You should leave this as it was.

loader/src/main/java/net/neoforged/neoforgespi/language/IModLanguageLoader.java
```diff
@@ -31,10 +37,23 @@
...
      */
     String name();
 
+    /**
+     * @return the version of this loader
+     */
+    default String version() {
```
Matyrobbrt, line 47: TBH, I'm inclined towards requiring locators to provide their version through whatever means they want.

Matyrobbrt (changes requested):

loader/src/main/java/net/neoforged/neoforgespi/language/IModLanguageLoader.java
thedarkcolour: I made it non-default, but I don't like how currently the same version code is in three different places.

loader/src/main/java/net/neoforged/fml/javafmlmod/FMLJavaModLanguageProvider.java
```diff
@@ -53,4 +58,15 @@ public void validate(IModFile file, Collection<ModContainer> loadedContainers, I
                     reporter.addIssue(issue);
                 });
     }
+
+    @Override
+    public String version() {
+        final Path lpPath;
```
Matyrobbrt, line 64: This should be in a helper or base class for all Neo lang providers.

loader/src/main/java/net/neoforged/neoforgespi/language/IModLanguageLoader.java
```diff
@@ -35,6 +35,11 @@ public interface IModLanguageLoader {
      */
     String name();
 
+    /**
+     * @return the version of this loader
```
Matyrobbrt, line 39:
```suggestion
     * {@return the version of this loader}
```

Matyrobbrt (changes requested):

loader/src/main/java/net/neoforged/neoforgespi/language/IModLanguageLoader.java
Matyrobbrt: That *is the correct* way of rendering inline return statements. They act as both a `@return` tag and as a `Returns (...).` description.

loader/src/main/java/net/neoforged/fml/loading/NeoLanguageLoader.java
```diff
@@ -0,0 +1,25 @@
...
+
+package net.neoforged.fml.loading;
+
+import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import net.neoforged.neoforgespi.language.IModLanguageLoader;
+
+public abstract class NeoLanguageLoader implements IModLanguageLoader {
```
Matyrobbrt, line 14: I would name this class BuiltIn rather than Neo.

Matyrobbrt (changes requested):

Matyrobbrt (approved):

Matyrobbrt (approved):

PR #139, opened 2024-05-16
Fix translation issues and introduce translation formatting for issues

shartte:
Translation args used in ModLoadingIssue errors have to start at index 3, while translations for warnings start at index 2. :-(

We had accidentally used translations in both error and warning messages, but due to this offset, we always have to use different translation keys.

Extend the tests for error conditions greatly and change them to assert on the english translation of those messages. That will then also test the actual interplay between translation args and formatting applied to those args in the default translation.

### DIFF
--- a/loader/src/test/java/net/neoforged/fml/loading/FMLLoaderTest.java
+++ b/loader/src/test/java/net/neoforged/fml/loading/FMLLoaderTest.java
@@ -446,7 +446,7 @@ class FMLLoaderTest extends LauncherTest {
             var e = assertThrows(ModLoadingException.class, () -> loadMods(launch("forgeclient")));
             assertThat(getTranslatedIssues(e.getIssues())).containsOnly(
                     "ERROR: testmod (testmod) is missing a feature it requires to run"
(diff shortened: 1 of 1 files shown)


loader/src/main/java/net/neoforged/fml/ModLoader.java
```diff
@@ -117,7 +117,7 @@ public static void gatherAndInitializeMods(final Executor syncExecutor, final Ex
         if (!failedBounds.isEmpty()) {
             LOGGER.fatal(CORE, "Failed to validate feature bounds for mods: {}", failedBounds);
             for (var fb : failedBounds) {
-                loadingIssues.add(ModLoadingIssue.error("fml.modloading.feature.missing", null, fb, ForgeFeature.featureValue(fb)).withAffectedMod(fb.modInfo()));
```
Technici4n, line 120: If you remove the `null` here you need to shift many indices in the translations. That's why it's still there (previously this was the "mod loading stage" which I removed; but I didn't want to adjust every single lang entry).

marchermans (dismissed):

marchermans (approved):

PR #144, opened 2024-05-24
Sort modded enum entries

agnor99:
https://github.com/neoforged/NeoForge/pull/970 uncovered that enum order for IExtensibleEnum enums is not guaranteed. As modders are encouraged they make their new enum values in their modconstructor multiple mods will try to add one at the same time. the create method is made synchronized, but whoever adds the values first still depends on timing. 

This adds the following code:
Field:
private  static List<EnumType> vanillaValues = null;
MethodCode:
```java
if (vanillaValues == null) {
    vanillaValues = Arrays.asList(values);
}
List<EnumType> modValues = new ArrayList<>(Arrays.asList(values));
modValues.removeAll(vanillaValues);
modValues.sort(ASMUtils.nameComparator());
modValues.addAll(0, vanillaValues); //now allOrderedValues
values = modValues.toArray(new EnumType[0]);
for (int i = vanillaValues.size(); i < values.length; i++) {
    ASMUtils.setOrdinal(values[i], i);
}
return newValue;
```

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/common/asm/RuntimeEnumExtender.java
+++ b/loader/src/main/java/net/neoforged/fml/common/asm/RuntimeEnumExtender.java
@@ -7,6 +7,10 @@ package net.neoforged.fml.common.asm;
 
 import com.mojang.logging.LogUtils;
 import cpw.mods.modlauncher.serviceapi.ILaunchPluginService;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


loader/src/main/java/net/neoforged/fml/common/asm/RuntimeEnumExtender.java
```diff
@@ -34,6 +38,21 @@ public class RuntimeEnumExtender implements ILaunchPluginService {
     private final String CLEAN_DESC = Type.getMethodDescriptor(Type.VOID_TYPE, Type.getType(Class.class));
     private final String NAME_DESC = Type.getMethodDescriptor(STRING);
     private final String EQUALS_DESC = Type.getMethodDescriptor(Type.BOOLEAN_TYPE, STRING);
+    private final Type LIST_TYPE = Type.getType(List.class);
+    private final Type ARRAY_LIST_TYPE = Type.getType(ArrayList.class);
+    private final String ARRAY_LIST_INIT_DESC = Type.getMethodDescriptor(Type.VOID_TYPE, Type.getType(Collection.class));
+    private final String VANILLA_FIELDS_SIGN = "Ljava/util/List<LMyExtensibleEnum;>";
```
RaymondBlaze, line 44:
```suggestion
    private final String VANILLA_FIELDS_SIGN = "Ljava/util/List<%s>;";
```
JVM doesn't really care about signatures, but lets make it right because we can :)

RaymondBlaze (changes requested):

loader/src/main/java/net/neoforged/fml/common/asm/RuntimeEnumExtender.java
agnor99: yeah, it doesn't really care, but can you point me to the place in the spec where %s is valid as I didn't see that

sciwhiz12: I believe this suggestion would be read alongside their other suggestion, which calls `formatted` on this string with the appropriate descriptor.

RaymondBlaze (approved):

embeddedt (changes requested):
I am not convinced that this is the way to go. Extending enums at runtime is already a hack and needing to do sorting like this only makes the hack worse. Additionally, I've had concerns about how the JVM deals with extending enums at runtime in general. It's already a known issue from 1.12 that if you extend an enum used in hot code (e.g. `DimensionType`), the JVM may have inlined the previous values, causing unexpected behavior.

Since we already have an ASM transformer in this area anyway, why don't we consider making the list of entries statically known at launch time (e.g. specified in JSON) and do the extension once when the class is loaded? This should be much safer as it's effectively equivalent to recompiling the enum.

agnor99:
are you sure that inlining can become a problem, when people are encouraged/expected to add to enums during modconstruction? not a lot of stuff is happening before that and I don't expect inlining to happen before that.

TheSilkMiner:
In 1.12, `DimensionType`s were added during mod construction, but that didn't stop the JVM from inlining. In fact, every access to an enum for whatever reason should be considered as a potential inlining trigger (assuming the enum is on a hot code path, that is).

RaymondBlaze:
> Since we already have an ASM transformer in this area anyway, why don't we consider making the list of entries statically known at launch time (e.g. specified in JSON) and do the extension once when the class is loaded? This should be much safer as it's effectively equivalent to recompiling the enum.

I personally agree with moving enum extension to an earlier timing, but JSON isn't really sufficient enough for the work, we still need some level of Java code to be ran in order to provide the constructor parameters.

PR #145, opened 2024-05-24
Be more specific about @Mod initializer order

Gaming32:
Common initializers are now always run before sided.

In this example, `MyMod` will always be run before `MyModClient`.
```java
@Mod(MyMod.MOD_ID)
class MyMod {
    public static final String MOD_ID = "mymod";
}

@Mod(value = MyMod.MOD_ID, dist = Dist.CLIENT)
class MyModClient {
}
```

### DIFF
--- a/loader/src/test/java/net/neoforged/fml/javafmlmod/FMLJavaModLanguageProviderTest.java
+++ b/loader/src/test/java/net/neoforged/fml/javafmlmod/FMLJavaModLanguageProviderTest.java
@@ -21,11 +21,13 @@ import org.junit.jupiter.api.Test;
 
 public class FMLJavaModLanguageProviderTest extends LauncherTest {
     // Allows us to capture events received by mods loaded in the game layer
-    public static List<FMLClientSetupEvent> EVENTS = new ArrayList<>();
+    public static final List<FMLClientSetupEvent> EVENTS = new ArrayList<>();
+    public static final List<String> MESSAGES = new ArrayList<>();
 
     @AfterEach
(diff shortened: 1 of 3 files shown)


Technici4n (changes requested):
I prefer the more implicit approach where side-specific entrypoints run after the common entrypoint.

Gaming32:
Maybe that would a good default, but the ability to specify an explicit ordering should still be an option in my opinion.

Technici4n (approved):

loader/src/test/java/net/neoforged/fml/javafmlmod/FMLJavaModLanguageProviderTest.java
```diff
@@ -107,6 +109,40 @@ public EntryPoint(net.neoforged.bus.api.IEventBus modEventBus) {
...
+                                }
+                            }
+                            """)
+                .compile();
+        }
+
+        var result = launch("forgeclient");
+        loadMods(result);
+
+        ModLoader.dispatchParallelEvent("test", Runnable::run, Runnable::run, () -> {}, FMLClientSetupEvent::new);
```
Technici4n, line 141: This shouldn't be necessary? `loadMods` will already construct the mods I think.

Gaming32: Copy+paste moment

Technici4n (approved):

Technici4n (approved):

PR #148, opened 2024-05-26
Data-driven enum extension

XFactHD:
This PR rewrites enum extension to append the new entries to the bytecode as if the class was compiled with them (for the most part at least) instead of inserting them at any random time with a lot of hacks.
To do this, enum entries are defined in a file and can either specify the necessary parameters as constant values in the file (only works for primitive types, strings and null) or provide a reference to a `public static final List<Object>` field in a class within the mod which can provide more complex parameter types.

The current implementation is functional but still needs improvements to error handling and sanity checking of input data, especially type and member signatures, and some test code removed. The provided unit test also needs further changes to run with a transforming class loader in order to actually test anything.

## Motivations

Enum extension currently has several issues:

- Order of mod-added enum constants is not guaranteed due to parallel mod loading and no enforced timing of the additions
- Additions after the enum's initialization requires substantial hacks to work around optimizations such as caching the `values` array and the map used for `Enum.valueOf()`. Certain JVM-level optimisations such as inlining of the values array may not be prevented by these hacks
- There is no guarantee at any time that an enum has all values it will ever have, risking issues with data structures like `EnumSet`s and `EnumMap`s
- Entry addition with arbitrary timing and the inability to properly specify certain parameters (mainly int ID parameters which should match the entry's ordinal) require substantial patches to make `Codec`s and `StreamCodec`s reasonably safe with the latter still falling victim to the missing guarantee of order

## Solutions

- The enum entries are sorted during loading to ensure consistent ordering between multiple launches
- Custom entries are added as if they were part of the class from the beginning, alleviating this issue entirely
- Same as above
- Consistent ordering, immediate presence of the values and special handling for enums with an int ID parameter that ensures its equality to the entry's ordinal removes the need for lazy initialization of `Codec`s and `StreamCodec`s and also makes the latter safe as long as the set of entries exactly matches between the two sides, which can be achieved with a configuration task in NeoForge

## Risks

The main risk with this approach is that modded values and their potential references to mod code will be loaded as soon as vanilla code loads a given enum. This puts a bigger emphasis on making sure that references to data which may only be available later is given the supplier treatment and also on the diligence of modders to make sure that loading classes which contain the parameter lists for enums with non-trivial constructors is safe.

## Details

To add entries to an enum, a mod needs to create a JSON file holding these additions. By default this file is expected at `META-INF/enumextender.json` but non-standard paths can be specified in the `neoforge.mods.toml` in the same way it can be done for mixin configs and AT files. The file has the following structure:

```json5
{
    "modid": "yourmodid",
    "entries": [
        {
            "enum": "net/minecraft/world/item/Rarity", // Class name of the enum
            // Name of the enum field to be added, must be prefixed with the mod ID specified above, otherwise it will be prefixed automatically
            "name": "NEOTESTS_CUSTOM",
            // Signature of the constructor as visible in the source code (omitting the hidden field name and ordinal params)
            "constructor": "(ILjava/lang/String;Ljava/util/function/UnaryOperator;)V",
            // "Complex" parameters provided as a reference to a field of type List<Object> in the given class
            "parameters": {
                "class": "net/neoforged/neoforge/debug/item/ItemTests",
                "field": "RARITY_ENUM_PARAMS"
            }
        },
        {
            "enum": "net/minecraft/world/item/ItemDisplayContext",
            "name": "NEOTESTS_HANGING",
            "constructor": "(ILjava/lang/String;Ljava/lang/String;)V",
            // Constant parameters provided directly.
            // This enum has an ID parameter, which enforces providing -1 in the case of constant parameters (complex parameters do not enforce this and instead just disregard this parameter in the list)
            "parameters": [ -1, "neotests:hanging", null ]
        },
        {
            "enum": "net/minecraft/world/entity/MobCategory",
            "name": "NEOTESTS_TEST",
            "constructor": "(Ljava/lang/String;IZZI)V",
            "parameters": [ "neotests:test", 1, true, true, 128 ]
        }
    ]
}
```

Enums with an int ID parameter get special treatment to make their ID the same as their ordinal if they are annotated with `@NumberedEnum`.
Enum constructors which must not be used by mods (i.e. `BiomeSpecialEffects.GrassColorModifier(String)`) can be annotated with `@BlacklistedConstructor` (I'd be happy about suggestions for a better name) to prevent them from being used.

The mod adding the new entry is then expected to retrieve it via `TheEnum.valueOf()` if the mod's code needs a reference to it.

Decompiling an enum after running the `RuntimeEnumExtender` provides the following output (see above for respective input):
```java
@NumberedEnum
public enum Rarity implements StringRepresentable, IExtensibleEnum {
    COMMON(0, "common", ChatFormatting.WHITE),
    UNCOMMON(1, "uncommon", ChatFormatting.YELLOW),
    RARE(2, "rare", ChatFormatting.AQUA),
    EPIC(3, "epic", ChatFormatting.LIGHT_PURPLE),
    NEOTESTS_CUSTOM(4, (String)ItemTests.RARITY_ENUM_PARAMS.get(1), (UnaryOperator)ItemTests.RARITY_ENUM_PARAMS.get(2));
    ...
}
```


agnor99:
would it be much more difficult to have a specialized object instead of List<Object> that can return the new enum value when queried for? It would remove a bit of the burden of looking up the values from the modders by name and removes the risk of modders having a typo in their enum names

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/EnumParameters.java
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) NeoForged and contributors
+ * SPDX-License-Identifier: LGPL-2.1-only
+ */
+
+package net.neoforged.fml.common.asm.enumextension;
+
+import java.util.List;
+import org.objectweb.asm.Type;
+
+public interface EnumParameters {
```
RaymondBlaze, line 11:
```suggestion
public sealed interface EnumParameters permits EnumParameters.Constant, EnumParameters.ListBased {
```
- Using sealed class here could optimize `RuntimeEnumExtender#loadConstructorParams` with enhanced switch.
- The subclasses names are a bit confusing, maybe `ListBacked` -> `FieldReference`?
- Supporting loading parameters from static methods should also be a choice.

XFactHD: Implemented the first two suggestions.

How would static methods work here? One method per parameter (more type safe but annoying for more than one or two parameters)? Single method with an index parameter (lacking type-safety like the list but simple to implement)?

Gaming32: It could also be expected to return a List, similar to the field, but potentially simpler to write.

XFactHD: I would prefer to keep the generated bytecode close to how the compiler would generate it, mainly because it means that decompiling a transformed class for debugging works better. This means that I would have to call that method once per parameter which would in turn require the modder to store the list in a field and return that, which completely defeats the point of using a method.

XFactHD:
@agnor99 I've implemented a possible approach for your suggestion. The one thing that bugs me about it is the necessity of forcing the class to be loaded in the value getter.

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/EnumProxy.java
```diff
@@ -0,0 +1,48 @@
...
+            } catch (ClassNotFoundException e) {
+                throw new RuntimeException(e);
+            }
+        }
+        return Objects.requireNonNull(enumValue, "Enum not initialized");
+    }
+
+    @ApiStatus.Internal
+    public void setValue(T value) {
+        this.enumValue = value;
```
Matyrobbrt, line 46: There's a small chance that this could run into concurrency issues. Though I'm not sure if locking in the get method is a good idea as it's a relatively hot path.

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/EnumPrototype.java
```diff
@@ -0,0 +1,109 @@
...
+                }
+                case "B" -> element.getAsByte();
+                case "S" -> element.getAsShort();
+                case "I" -> element.getAsInt();
+                case "F" -> element.getAsFloat();
+                case "J" -> element.getAsLong();
+                case "D" -> element.getAsDouble();
+                case "Ljava/lang/String;" -> element.isJsonNull() ? null : element.getAsString();
+                default -> {
+                    if (element.isJsonNull() || element.getAsString().equals("null")) {
```
Matyrobbrt, line 98: I'd remove the ability of it being a string with `null` as its value, feels like too much magic when json already has null values.

Matyrobbrt (changes requested):

agnor99:
would it be possible to get access to the list of vanilla or modded values or an int for the first modded value, which could be used to fix neoforged/Neoforge#150 in an easier way and to shrink the payload size in neoforged/NeoForge/pull/970?

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/EnumPrototype.java
```diff
@@ -0,0 +1,109 @@
...
+            String owner = json.get("modid").getAsString();
+
+            JsonArray entries = json.getAsJsonArray("entries");
+            List<EnumPrototype> prototypes = new ArrayList<>(entries.size());
+            for (JsonElement entry : entries) {
+                JsonObject entryObj = entry.getAsJsonObject();
+
+                String enumName = entryObj.get("enum").getAsString();
+
+                String fieldName = entryObj.get("name").getAsString();
```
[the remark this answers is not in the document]
XFactHD: Validation of all the descriptors and this field is something I still need to look into as mentioned in the description.

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/IExtensibleEnum.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * To be implemented on vanilla enums that should be enhanced with ASM to be
+ * extensible. If this is implemented on a class, the class must define a static
+ * method called {@code isExtended()} which takes zero args and returns a boolean.
+ * By default, the method should throw to make sure the enum was handled by the transformer.
+ * <p>
+ *
+ * <pre>
+ * public static boolean isExtended() {
```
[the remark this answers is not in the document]
XFactHD: The goal is to use this method in neoforged/NeoForge#970 to check whether the enum contains modded values or not. It is not automatically used by anything.

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/EnumProxy.java
XFactHD: Shouldn't this be prevented by the explicit classload in the getter? I thought there was some kind of lock involved to prevent a class from being initialized on multiple threads.

embeddedt: > I thought there was some kind of lock involved to prevent a class from being initialized on multiple threads.

That is correct; for our purposes, `<clinit>` behaves as though synchronized to the class type. It'd probably be wise to mark `enumValue` as `volatile` though.

XFactHD:
@agnor99 The idea behind the static `isExtended()` method was to cover that need. It returns true if any modded values were added.

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/EnumParameters.java
RaymondBlaze:
```java
public static Object getParameter(int index) {
    return switch (index) {
        case 0 -> someObj;
        //...
    }
}

//With generics
public static <T> T getParameter(int index, Class<T> type) {
    return switch (index) {
        case 0 -> type.cast(someObj);
        //...
    }
}
```
There's no need to return a list when using a static method, the above example should be sufficient enough.
Additionally, the method could be made generic and pass in a `Class<T>` parameter to enforce the cast being done in the modder's method first, so any `ClassCastException` will not be thrown from the enum class.

RaymondBlaze:
A better alternative name for `@BlacklistedConstructor` would be `@ReservedConstructor`.

RaymondBlaze:
> The idea behind the static `isExtended()` method was to cover that need. It returns true if any modded values were added.

I believe what agnor meant was that the modded values' id, name and owning mod are needed in the handshake to verify the exact mismatch.
The fact that the enum is extended is only sufficient enough for a NeoForge <-> Vanilla connection, in NeoForge <-> NeoForge connection, mods could be adding different enum instances and that would need further verification.
The `boolean isExtended()` method could just be replaced by "int getExtensions" where the implementation would be:
```java
int getExtensions() {
    return values().length - <the count of original values>;
}
```

agnor99:
that was not the point I was making. The problem with only knowing that it was extended is not enough to fix the RecipeBookSettings#read part when trying to connect with a modded client that has extended the RecipeBookType enum to a vanillalike serverlobby as we need to know when to stop reading values. For that we need to know how many original values were in there originally as that is how many a vanillalike server is going to send.
The benefit to also having that information is that we can then throw away the vanilla values and don't have to send them in the enum verify payload because vanillavalues can be expected to exist.

XFactHD:
Not having to send vanilla values is a good point. Though my preference would then be to return an "extension info" record that has an extension flag, the last vanilla ordinal and either the added count or the total count.

XFactHD:
@agnor99 I've replaced the extension flag getter with a getter returning a record that contains a flag for wether the enum is extended, how many vanilla values it contains, how many total values it contains after extension and to which degree it needs to be checked for networking compatibility.

agnor99:
lgtm

Fuzss:
Would it be possible to track what mod a custom enum constant was added by, e.g. via a namespace?

I'm thinking like enforcing valid resource locations for `name`, like `neoforge:custom`, which then internally could be transformed to `NEOFORGE_CUSTOM` for the added enum constant. Just having the mod id be some part of the new constant would be the simplest way to go about it I guess. Just my idea, maybe there are other approaches for tracking the origin? Alternatively maybe a method in IExtensibleEnum` returning a string namespace?

I've worked a lot with extensible enums like the old `EnchantmentCategory` in the past, and with a lot of mods adding custom values working out issues with specific values has always been difficult since there is no way to find a constants origin besides removing mods and observing when the problematic value vanishes.

XFactHD:
> Would it be possible to track what mod a custom enum constant was added by, e.g. via a namespace?
> 
> I'm thinking like enforcing valid resource locations for name, like neoforge:custom [...]

@Fuzss I've added validation on name parameters to ensure they are namespaced. Having the enum entries themselves keep track of the mod that added them is not feasible outside of what the constant's name (which is validated for being prefixed with the mod ID) and, if present, a namespaced name parameter provide. However, the `RuntimeEnumExtender` keeps the "enum prototypes" around, so you could in theory use that to check for the source of a constant via the debugger.

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/EnumPrototype.java
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                } else {
+                    throw new IllegalArgumentException(String.format(
+                            Locale.ROOT,
+                            "Unexpected parameter declaration '%s' for field '%s' in enum '%s' specified by mod '%s'",
+                            paramElem,
+                            fieldName,
+                            enumName,
+                            modId));
+                }
```
Matyrobbrt, line 126: There's a bit of exception duplication here.

tmvkrpxl0:
Just a though, would it be feasible to have this feature in the mixin itself?

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/IndexedEnum.java
```diff
@@ -0,0 +1,20 @@
...
+package net.neoforged.fml.common.asm.enumextension;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+
+/**
+ * Specifies that the annotated enum has an int ID parameter which must match the enum constant's ordinal
+ */
+@Target(ElementType.TYPE)
+public @interface IndexedEnum {
```
Matyrobbrt, line 15: Would it be an idea to have this as a parameter level annotation in the ctor?

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/NamedEnum.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+
+/**
+ * Specifies that the annotated enum has a string name parameter which must be prefixed with the mod ID
+ * of the mod adding a given enum constant
+ */
+@Target(ElementType.TYPE)
+public @interface NamedEnum {
```
Matyrobbrt, line 16: Same as indexed enum

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/RuntimeEnumExtender.java
```diff
@@ -0,0 +1,435 @@
...
+    private static volatile Map<String, List<EnumPrototype>> prototypes = Map.of();
+
+    @Override
+    public String name() {
+        return "runtime_enum_extender";
+    }
+
+    @Override
+    public EnumSet<Phase> handlesClass(Type classType, boolean isEmpty) {
+        return isEmpty ? NAY : YAY;
```
Matyrobbrt, line 72: Should we limit it to vanilla classes, or should we allow other mods to apply this on their enums (or on other mods' enums through a mixin)

loader/src/main/java/net/neoforged/fml/loading/LoadingModList.java
```diff
@@ -5,31 +5,37 @@
...
 import net.neoforged.fml.loading.modscan.BackgroundScanHandler;
 import net.neoforged.neoforgespi.language.IModInfo;
+import org.slf4j.Logger;
 
 /**
  * Master list of all mods <em>in the loading context. This class cannot refer outside the
  * loading package</em>
  */
 public class LoadingModList {
+    private static final Logger LOGGER = LogUtils.getLogger();
```
Matyrobbrt, line 38: Is this used?

loader/src/main/java/net/neoforged/fml/loading/LoadingModList.java
```diff
@@ -87,6 +93,22 @@ public void addAccessTransformers() {
                 .forEach(mod -> mod.getAccessTransformers().forEach(path -> FMLLoader.addAccessTransformer(path, mod)));
     }
 
+    public void addEnumExtenders() {
+        Map<IModInfo, Path> pathPerMod = new HashMap<>();
```
Matyrobbrt, line 97: Is there value in allowing a mod to have multiple extenders?

loader/src/main/resources/lang/en_us.json
```diff
@@ -56,5 +56,20 @@
...
+  "fml.modloading.enumextender.argument.reference.unexpected_decl": "Unexpected reference parameter declaration {3} for field {4} in enum {5} specified by mod {0,modinfo,id}",
+  "fml.modloading.enumextender.argument.unexpected_decl": "Unexpected parameter declaration {3} for field {4} in enum {5} specified by mod {0,modinfo,id}",
+  "fml.modloading.enumextender.loading_error": "Failed to load enum extension data at {3} in mod {0,modinfo,id}: {2,exc,msg}",
+  "fml.modloading.enumextender.argument.constant.count_mismatch": "Parameter count {3} does not match argument count {4} of constructor {5} for field {6} in enum {7} specified by mod {0,modinfo,id}",
+  "fml.modloading.enumextender.argument.constant.invalid_char": "Invalid character {3} at parameter index {4} for field {5} in enum {6} specified by mod {0,modinfo,id}",
+  "fml.modloading.enumextender.argument.constant.unsupported_type": "Unsupported immediate argument type {3} at parameter index {4} for field {5} in enum {6} specified by mod {0,modinfo,id}",
+  "fml.modloading.enumextender.duplicate": "Found duplicate field {3} for enum {4} provided by mods {5} and {6}"
```
Matyrobbrt, line 74: Could do with some commas surrounding `provided by mods` or `specified by mod`

Matyrobbrt (changes requested):

Matyrobbrt:
> Just a though, would it be feasible to have this feature in the mixin itself?

Hmm? If you mean that this feature should be in mixin, then no, the main reason this is chosen over the current system is for networked enums, so it's quite closely tied to minecraft.

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/IndexedEnum.java
Matyrobbrt: Disregard, this would increase the patch length significantly, making it unreadable.

loader/src/main/resources/lang/en_us.json
Matyrobbrt:
```suggestion
  "fml.modloading.enumextender.file_not_found": "Enum extender file {3}, provided by mod {0,modinfo,id}, does not exist",
  "fml.modloading.enumextender.invalid_enum_name": "Enum {3}, specified by mod {0,modinfo,id}, is not a valid class descriptor",
  "fml.modloading.enumextender.field_name.missing_prefix": "Enum constant name {3} for enum {4}, specified by mod {0,modinfo,id}, must be prefixed with the mod ID",
  "fml.modloading.enumextender.field_name.invalid": "Enum constant name {3} for enum {4}, specified by mod {0,modinfo,id}, is invalid",
  "fml.modloading.enumextender.invalid_constructor": "Constructor {3} for enum {4}, specified by mod {0,modinfo,id}, is not a valid constructor descriptor",
  "fml.modloading.enumextender.argument.reference.invalid_src_class": "Parameter source class {3} for field {4} in enum {5}, specified by mod {0,modinfo,id}, is not a valid class descriptor",
  "fml.modloading.enumextender.argument.reference.invalid_src_method": "Parameter source method {3} for field {4} in enum {5}, specified by mod {0,modinfo,id}, is not a valid method name",
  "fml.modloading.enumextender.argument.reference.invalid_src_field": "Parameter source field {3} for field {4} in enum {5}, specified by mod {0,modinfo,id}, is not a valid field name",
  "fml.modloading.enumextender.argument.reference.unexpected_decl": "Unexpected reference parameter declaration {3} for field {4} in enum {5}, specified by mod {0,modinfo,id}",
  "fml.modloading.enumextender.argument.unexpected_decl": "Unexpected parameter declaration {3} for field {4} in enum {5}, specified by mod {0,modinfo,id}",
  "fml.modloading.enumextender.loading_error": "Failed to load enum extension data at {3}, in mod {0,modinfo,id}: {2,exc,msg}",
  "fml.modloading.enumextender.argument.constant.count_mismatch": "Parameter count {3} does not match argument count {4} of constructor {5}, for field {6}, in enum {7}, specified by mod {0,modinfo,id}",
  "fml.modloading.enumextender.argument.constant.invalid_char": "Invalid character {3} at parameter index {4}, for field {5}, in enum {6}, specified by mod {0,modinfo,id}",
  "fml.modloading.enumextender.argument.constant.unsupported_type": "Unsupported immediate argument type {3}, at parameter index {4}, for field {5}, in enum {6}, specified by mod {0,modinfo,id}",
  "fml.modloading.enumextender.duplicate": "Found duplicate field {3}, for enum {4}, provided by mods {5} and {6}"
```

RaymondBlaze:
> Just a though, would it be feasible to have this feature in the mixin itself?

Not really, though achieving this feature on with pure (but a bit hacky) mixin implementation is possible (I've actually tried with even enum subclassing included) and such implementation do has benefits as being multiloader friendly, this PR focus more on vanilla (and especially networked) enums, and do note that `IExtensibleEnum` is internal for FML/NeoForge uses only, so there's no point to move this to something with a bigger scope like Mixin.

By the way, I wonder how will this (and enum network checks) work with legacy-ish Mixin based enum extensions in case multiloader mods are still using that implementation for keeping things on common.

loader/src/main/java/net/neoforged/fml/loading/LoadingModList.java
XFactHD: Not sure, to be honest

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/RuntimeEnumExtender.java
XFactHD: Moving the extender to run before mixin implicitly makes it impossible to make additional enums, vanilla and modded, extensible via mixin

XFactHD:
> By the way, I wonder how will this (and enum network checks) work with legacy-ish Mixin based enum extensions in case multiloader mods are still using that implementation for keeping things on common.

The extender now runs before mixin, which ensures that entries added via mixin always end up after ones added via this system (assuming people use the actual last entry and not just the last source-visible constant to determine the ordinal of their mixined entry) such that they at least can't break the sorting guarantee and network checks provided by this system.

Matyrobbrt (approved):

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/RuntimeEnumExtender.java
```diff
@@ -0,0 +1,435 @@
...
+    }
+
+    @Override
+    public EnumSet<Phase> handlesClass(Type classType, boolean isEmpty) {
+        return isEmpty || !prototypes.containsKey(classType.getInternalName()) ? NAY : YAY;
+    }
+
+    @Override
+    public boolean processClass(final Phase phase, final ClassNode classNode, final Type classType) {
+        if ((classNode.access & Opcodes.ACC_ENUM) == 0 || !classNode.interfaces.contains(MARKER_IFACE.getInternalName())) {
```
Technici4n, line 77: Shouldn't it be an error to try to extend a non-enum class?

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/RuntimeEnumExtender.java
```diff
@@ -0,0 +1,435 @@
...
+    private static final Type EXT_INFO = Type.getType(ExtensionInfo.class);
+    private static final String EXT_INFO_GETTER_DESC = Type.getMethodDescriptor(EXT_INFO);
+    private static final String EXT_INFO_CTOR_DESC = Type.getMethodDescriptor(
+            Type.VOID_TYPE, Type.BOOLEAN_TYPE, Type.INT_TYPE, Type.INT_TYPE, NET_CHECK);
+    private static final Type NETWORKED_ANNOTATION = Type.getType(NetworkedEnum.class);
+    private static final Type EXTENDER = Type.getType(RuntimeEnumExtender.class);
+    private static final Type ARRAYS = Type.getType("Ljava/util/Arrays;");
+    private static final int ENUM_FLAGS = Opcodes.ACC_PUBLIC | Opcodes.ACC_STATIC | Opcodes.ACC_FINAL | Opcodes.ACC_ENUM;
+    private static final int EXT_INFO_FLAGS = Opcodes.ACC_PRIVATE | Opcodes.ACC_STATIC | Opcodes.ACC_FINAL;
+    private static volatile Map<String, List<EnumPrototype>> prototypes = Map.of();
```
Technici4n, line 63: `volatile` is not necessary since this happens in a single-threaded phase. Future uses of multi-threading should establish an appropriate `happens-before` relationship.

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/RuntimeEnumExtender.java
```diff
@@ -0,0 +1,435 @@
...
+            return false;
+        }
+
+        List<EnumPrototype> protos = prototypes.getOrDefault(classType.getInternalName(), List.of());
+        if (protos.isEmpty()) {
+            return false;
+        }
+
+        MethodNode clinit = findMethod(classNode, mth -> mth.name.equals("<clinit>"));
+        MethodNode $values = findMethod(classNode, mth -> mth.name.equals("$values"));
```
Technici4n, line 87: Question: Does ECJ compile enums in the same way? This matters for neodev in Eclipse, and the upcoming moddev plugin which can use ECJ to recompile (because it's faster).

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/IExtensibleEnum.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * {@snippet :
+ * public static ExtensionInfo getExtensionInfo() {
+ *     return ExtensionInfo.nonExtended(TheEnum.class);
+ * }
+ * }
+ *
+ * The method contents will be replaced with ASM at runtime to return information about the extension of the enum
+ */
+public interface IExtensibleEnum {}
```
Technici4n, line 22: It would maybe be nice to add an example of how to access added entries in the javadoc here. To be fair, we should have centralized documentation somewhere because it's not trivial.

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/EnumProxy.java
```diff
@@ -0,0 +1,54 @@
...
+        return parameters.get(idx);
+    }
+
+    /**
+     * {@return the created enum constant, ensuring the containing enum is loaded if the value was not set yet}
+     */
+    public T getValue() {
+        if (enumValue == null) {
+            try {
+                var loader = StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE).getCallerClass().getClassLoader();
```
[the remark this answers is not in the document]
XFactHD: If the value is still null, then it's not loaded yet. For example, when you call this in a static field initializer like `private static final Rarity MY_RARITY = AnotherClass.RARITY_ENUM_PROXY.getValue()`, then the `Objects.requireNonNull()` below blows up, presumably because only the `PUTSTATIC` after the invocation of `getValue()` actually causes a classload.

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/RuntimeEnumExtender.java
XFactHD: That's a fair point, yeah. I'll add a throw if the target is not an enum.

loader/src/main/java/net/neoforged/fml/common/asm/ListGeneratorAdapter.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package net.neoforged.fml.common.asm;
+
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.commons.GeneratorAdapter;
+import org.objectweb.asm.tree.InsnList;
+import org.objectweb.asm.tree.MethodNode;
+
+public class ListGeneratorAdapter extends GeneratorAdapter {
```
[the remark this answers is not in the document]
XFactHD: In theory yes, if I move it into the enumextension package. Though, this seems like something that's generally useful for transformers (including mod-provided coremods)

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/RuntimeEnumExtender.java
```diff
@@ -0,0 +1,492 @@
...
+        ListGeneratorAdapter clinitTailGenerator = new ListGeneratorAdapter(new InsnList());
+        buildExtensionInfo(classNode, classType, clinitTailGenerator, infoField, vanillaEntryCount, protos.size());
+        returnValuesToExtender(classType, clinitTailGenerator, protos, enumEntries);
+        AbstractInsnNode clinitRetNode = ASMAPI.findFirstInstructionBefore(clinit, Opcodes.RETURN, clinit.instructions.size() - 1);
+        clinit.instructions.insertBefore(clinitRetNode, clinitTailGenerator.insnList);
+        classNode.fields.addAll(vanillaEntryCount, enumEntries);
+
+        ListGeneratorAdapter appendValuesGenerator = new ListGeneratorAdapter(new InsnList());
+        appendValuesArray(classType, appendValuesGenerator, enumEntries);
+        if ($valuesOpt.isPresent()) { // javac
```
Matyrobbrt, line 134: I would add a boolean instead of using isPresent every time.

Matyrobbrt (changes requested):

Matyrobbrt (approved):

PR #150, opened 2024-05-27
Add STARTUP configs for early config loading/usage

TelepathicGrunt:
I think this is all that is needed. JUnit tests ran so I didn't completely kill the build lol

See https://github.com/neoforged/NeoForge/issues/723


Technici4n:
I'm happy to test this once there is a corresponding NF PR. :wink:

TelepathicGrunt:
@Technici4n what needs to happen on the Neo PR? Just a test mod there to test it? How would I have the Neo pr build against this fml pr?

Technici4n:
You can use PR publishing for that. 😉

Check https://github.com/neoforged/FancyModLoader/pull/150#issuecomment-2134028746 once it's finished publishing.

Technici4n (dismissed):

loader/src/main/java/net/neoforged/fml/config/ConfigTracker.java
```diff
@@ -76,7 +76,7 @@ private Path resolveBasePath(ModConfig config, Path configBasePath, @Nullable Pa
         return configBasePath;
     }
 
-    private void openConfig(final ModConfig config, final Path configBasePath, @Nullable Path configOverrideBasePath) {
+    public void openConfig(final ModConfig config, final Path configBasePath, @Nullable Path configOverrideBasePath) {
```
Matyrobbrt, line 79: You should either mark this as internal, or the entire class.

loader/src/main/java/net/neoforged/fml/config/ModConfig.java
```diff
@@ -94,19 +94,24 @@ public enum Type {
...
-        SERVER;
+        SERVER,
+        /**
+         * Startup config is for configuration that needs to run as early as possible.
+         * Only loaded right after the config are registered to FML.
+         * Please know what you are doing as using these configs to enable/disable registration and other specific stuff
+         * can cause clients to have issues connecting to servers with a different Startup config value.
+         * Stored in the global config directory.
+         * Not synced.
+         * Suffix is "-startup" by default.
```
Matyrobbrt, line 112:
```suggestion
         * Startup configs are for configurations that need to run as early as possible.
         * Loaded as soon as the config is registered to FML.
         * Please be aware when using them, as using these configs to enable/disable registration and anything that must be present on both sides
         * can cause clients to have issues connecting to servers with different config values.
         * Stored in the global config directory.
         * Not synced.
         * Suffix is "-startup" by default.
```

Matyrobbrt (changes requested):

Matyrobbrt (approved):

PR #152, opened 2024-06-01
Implement auto bus detection in EBS

### LINKED ISSUE #140: Make `EventBusSubscriber` automatically look up and dispatch mod bus events as necessary
`@EventBusSubscriber` has the capacity to be provided a `modid` argument, which will be used to lookup the mod bus for the target class.  We should leverage this functionality to permit automatic dispatch of mod bus and non-mod bus events to the correct bus within a single class as long as the modid is present.  This would also mean we can remove the `Bus` argument and enum.

We should also consider adding a non-static method for doing so (though this may need to be in NeoForge, and not here).  Potentially `NeoForge.registerEventSubscriber(modid, Object / Class)`, mimicking this new functionality but with `IEventBus#register` semantics.

The current paradigm requires separation of mod bus and non-mod bus events at the `class` level, ex:
```java
@EventBusSubscriber(modid = "mymod", Bus = Bus.MOD)
public class ModEvents {

    @SubscribeEvent
    public static void setup(FMLCommonSetupEvent e) {
        // Do setup tasks
    }

    @EventBusSubscriber
    public static class ForgeBusEvents {
        
        @SubscribeEvent
        public static void tick(LevelTickEvent.Pre e) {
            // Do something on tick
        }
    }
}
```
This approach does not provide meaningful separation (there is no reason that these must be separated at the `class` level, aside from the current EBS implementation), and unifying these events in the same class can readily be achieved via `IEventBus#addListener`.

Further, implementing such a feature will reduce the need to answer support questions about registering events to the correct event bus.

Matyrobbrt:
Closes #140


loader/src/main/java/net/neoforged/fml/javafmlmod/AutomaticEventSubscriber.java
```diff
@@ -42,22 +48,47 @@ public static void inject(final ModContainer mod, final ModFileScanData scanData
...
+
+                                var type = method.getParameterTypes()[0];
+                                if (IModBusEvent.class.isAssignableFrom(type)) {
+                                    hasMod = true;
+                                } else if (Event.class.isAssignableFrom(type)) {
+                                    hasGame = true;
+                                }
+                            }
+
+                            if (hasGame && hasMod) {
```
marchermans, line 77: Although I understand the idea, but why do we limit this to one bus per listener?

marchermans: Oke perfect, was just curious.  Thanks for the explanation. Nice work!

marchermans (approved):

Technici4n (changes requested):
I don't like this. It doesn't make life easier for beginners because they still need to understand the difference between the two kinds of busses. And it doesn't really make life easier for more experienced developers because typing the kind of bus you want is already trivial. This just add another way of doing the same thing, which is generally a sign of incoherent design.

loader/src/main/java/net/neoforged/fml/javafmlmod/AutomaticEventSubscriber.java
RaymondBlaze: > This approach does not provide meaningful separation (there is no reason that these must be separated at the `class` level, aside from the current EBS implementation), and unifying these events in the same class can readily be achieved via `IEventBus#addListener`.

Despite the explained "limitation", leaving a class level separation here does not fix #140's main expectations.

Moving further to change the EBS implementation would be better than an incomplete solution imo.

Matyrobbrt: There is no raw variant of `addListener` that would support that use case, and reflection based consumers on the neo side is not acceptable, both because of performance concerns and because it will make unregistering those listeners impossible. In any case, I've decided to not proceed with this PR further because more magic will increase confusion. 

Instead, we should make the bus parameter required and the mod id default to first in the file (instead of all).

PR #155, opened 2024-06-02
Replace use of "extension" with explicit method on RuntimeDistCleaner

shartte:
This removes the need for unchecked casts in `getExtension()`.


loader/src/main/java/net/neoforged/fml/common/asm/RuntimeDistCleaner.java
```diff
@@ -128,22 +127,18 @@ private static List<AnnotationNode> unpack(final List<AnnotationNode> anns) {
...
     }
 
-    @SuppressWarnings("unchecked")
-    @Override
-    public Consumer<Dist> getExtension() {
-        return (s) -> {
-            DIST = s.name();
-            LOGGER.debug(DISTXFORM, "Configuring for Dist {}", DIST);
-        };
+    public static void setDistribution(Dist dist) {
```
XFactHD, line 139: This method should either be non-static or be called statically instead of on an instance of this class

XFactHD (changes requested):

XFactHD (approved):

marchermans:
I understand that this breaks binary compat, but in my eyes we can merge this. What do you guys think?

PR #160, opened 2024-06-04
Don't add modFolders as mods directly. Rather use them to group existing classpath entries

shartte:
**Note:** Breaking change, since users might currently set Source Sets in NG which are ultimately not on the classpath (this leads to debugging issues in IJ, but it might still be the case). We might want to hide this behind a flag, or use a different system property for it.

This passes control over which moddev mods get loaded to the classpath used when setting up the JVM, which will usually be Gradle or an IDE. This setup allows Gradle plugins to always pass all "potential" folders where compiled mod classes could be, and let FML sort out which ones are actually present on the classpath. Example: add the IntelliJ and Gradle Folders for IJ runs since we cannot know which build-tool is used by IJ. (same for Eclipse)

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/locators/UserdevLocator.java
+++ b/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/locators/UserdevLocator.java
@@ -5,12 +5,6 @@
 
 package net.neoforged.fml.loading.moddiscovery.locators;
 
-import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


loader/src/main/java/net/neoforged/fml/loading/moddiscovery/locators/UserdevLocator.java
```diff
@@ -28,10 +36,29 @@ public UserdevLocator(Map<String, List<Path>> modFolders) {
 
     @Override
     public void findCandidates(ILaunchContext context, IDiscoveryPipeline pipeline) {
+
+        // Index the JVM classpath
+        var entriesOnClasspath = new HashSet<>();
+        for (String entry : System.getProperty("java.class.path").split(Pattern.quote(File.pathSeparator))) {
+            try {
+                var path = Paths.get(entry);
```
marchermans, line 44: Would it not make sense to check if this is a directory?

PR #163, opened 2024-06-12
Use TransformingClassLoader in Tests and Write Data-Driven Enum Extension Tests

shartte:
Change config key from `enumExtender` to `enumExtensions` since this is still the chance to have it named according to what the setting does.
The JSON file only contains extensions to enums in a declarative manner, which means `enumExtensions` is a better name.


loader/src/main/java/net/neoforged/fml/loading/LoadingModList.java
```diff
@@ -92,9 +92,9 @@ public void addEnumExtenders() {
         modFiles.stream()
                 .map(ModFileInfo::getMods)
                 .flatMap(List::stream)
-                .forEach(mod -> mod.getConfig().<String>getConfigElement("enumExtender").ifPresent(file -> {
+                .forEach(mod -> mod.getConfig().<String>getConfigElement("enumExtensions").ifPresent(file -> {
```
marchermans, line 95: Why was this changed?

loader/src/main/java/net/neoforged/fml/common/asm/enumextension/RuntimeEnumExtender.java
```diff
@@ -452,7 +452,7 @@ public static void loadEnumPrototypes(Map<IModInfo, Path> paths) {
                 .reduce(
                         new HashMap<>(),
                         (map, proto) -> {
-                            map.computeIfAbsent(proto.enumName(), $ -> new ArrayList<>()).add(proto);
```
Matyrobbrt, line 455: You really hate money :P

loader/src/main/java/net/neoforged/fml/loading/LoadingModList.java
```diff
@@ -92,9 +92,9 @@ public void addEnumExtenders() {
...
                     Path path = mod.getOwningFile().getFile().findResource(file);
-                    if (Files.notExists(path)) {
+                    if (!Files.isRegularFile(path)) {
                         ModLoader.addLoadingIssue(ModLoadingIssue.error("fml.modloading.enumextender.file_not_found", path).withAffectedMod(mod));
```
Matyrobbrt, line 98: I feel like it should have a better error message if for some reason it's a directory.

Matyrobbrt: Very well then

Matyrobbrt (approved):

PR #167, opened 2024-06-22
Refactor ModLoadingIssue Warning/Error Translations

shartte:
Change the translation logic for mod loading issues warnings/errors to:

- Implicit arguments start at index 100
- Both warnings and errors get the same implicit arguments
- All translation keys for mod loading issues have to start with `fml.modloadingissue.` to make it easy to know when the implicit args are available
- Mods can still use their own translation keys for issues

Note that this is easier to review in the commit view. The updates to translation keys were done with a Python script, only the first contains manual code changes.


Matyrobbrt:
What's the motivation behind the modloading -> modloadingissue change?

HenryLoenwind:
Wouldn't `fml.modloading.issue` be a better fit for the naming scheme?

Matyrobbrt (approved):

PR #169, opened 2024-06-24
Allow multiple copies of a class on the CP and pick the JAR containing the first.

shartte:
This allows unusual setups where through cross-project or cross-source-set dependencies, IDEs put MC on the classpath multiple times.

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/util/DevEnvUtils.java
+++ b/loader/src/main/java/net/neoforged/fml/util/DevEnvUtils.java
@@ -13,13 +13,19 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


Technici4n (changes requested):
I'd prefer the MC jar override to be manual via a system prop. Is that unreasonable? That would better match NG's logic.

marchermans:
> I'd prefer the MC jar override to be manual via a system prop. Is that unreasonable? That would better match NG's logic.

Yes and no.
This follows NGs logic much better, since NGs logic relies on classpath ordering specifications as defined by Oracle for the JVM `--classpath` parameter.
The addition of NGs removal and subsequent further modification of the classpath in general is caused by the fact that with duplicates it won't start at all.

marchermans:
@Technici4n What is the state on this?
We still need this

Technici4n:
In case this got lost: my preference is still to pass the desired MC jar and client-extra as a system prop or something like that.

PR #170, opened 2024-06-24
Better error reporting in dev

shartte:
- This makes use of the changes in #167 to be able to freely convert a translation between error/warning
- It now reports the full set of incompatibility issues as errors when in dev, instead of falling back to a generic "brokenfile" error as before
- Added tests for more cases
- Made it so that an unknown FMLModType is now reported as a normal incompatibility, using the following adjustment:
  - If a ModFileReader throws a ModLoadingException, the contained issues are recorded as incompatibilities and will be reported, if no other reader successfully reads the file. The reporting level will be inherited (warning in prod, error in dev)
  - Any other exception continues to be reported as a loading error


loader/src/main/resources/lang/en_us.json
```diff
@@ -26,16 +26,18 @@
...
+  "fml.modloadingissue.brokenfile.unknownfmlmodtype": "File {101} has an unrecognized FML mod-type: ''{0}''",
+  "fml.modloadingissue.brokenfile.oldforge": "File {101} is for an old version of Minecraft Forge and cannot be loaded",
+  "fml.modloadingissue.brokenfile.minecraft_forge": "File {101} is for Minecraft Forge or an older version of NeoForge, and cannot be loaded",
+  "fml.modloadingissue.brokenfile.liteloader": "File {101} is a LiteLoader mod and cannot be loaded",
+  "fml.modloadingissue.brokenfile.fabric": "File {101} is a Fabric mod and cannot be loaded",
+  "fml.modloadingissue.brokenfile.quilt": "File {101} is a Quilt mod and cannot be loaded",
+  "fml.modloadingissue.brokenfile.optifine": "File {101} is an incompatible version of OptiFine",
+  "fml.modloadingissue.brokenfile.bukkit": "File {101} is a Bukkit or Bukkit-implementor (Spigot, Paper, etc.) plugin and cannot be loaded",
+  "fml.modloadingissue.brokenfile.invalidzip": "File {101} is not a jar file",
+  "fml.modloadingissue.brokenresources": "File {101} failed to load a valid ResourcePackInfo",
```
Matyrobbrt, line 40: Why are you not updating the indexes 
 everywhere?

Matyrobbrt (changes requested):

Matyrobbrt (approved):

PR #171, opened 2024-06-24
Use ByteBuddy in unit tests to avoid having to add add-opens to unit tests

shartte:
What it says on the tin :-)

We use the ByteBuddy agent implicitly already through Mockito.

### DIFF
--- a/loader/src/test/java/net/neoforged/fml/loading/LauncherTest.java
+++ b/loader/src/test/java/net/neoforged/fml/loading/LauncherTest.java
@@ -78,7 +78,7 @@ public abstract class LauncherTest {
     protected TransformingClassLoader gameClassLoader;
 
     @BeforeAll
-    public static void ensureAddOpensForUnionFs() {
+    static void ensureAddOpensForUnionFs() {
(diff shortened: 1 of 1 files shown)


loader/src/test/java/net/neoforged/fml/loading/LauncherTest.java
```diff
@@ -76,6 +78,17 @@ public abstract class LauncherTest {
 
     @BeforeEach
     void setUp() throws Exception {
+        // We abuse the ByteBuddy agent that Mockito also uses to open java.lang to UnionFS
```
Matyrobbrt, line 81: You can move this into a beforeall

Matyrobbrt (changes requested):

Matyrobbrt (approved):

SquidDev:
Apologies if you're already aware, but there's plans to disable dynamic loading of agents (and so `ByteBuddyAgent.install`) in the future — see [JEP 451](https://openjdk.org/jeps/451). This definitely doesn't matter right now, as NeoForged is pinned to Java 21, but worth bearing in mind for the future!

PR #172, opened 2024-06-26
Update NightConfig and cleanup config code

Technici4n:
Change list:
- Update to the latest NightConfig. Configs are now threadsafe, and reloading works correctly with Notepad++.
- Remove internal `ConfigConfig` tech debt.
- Redesign `IConfigSpec` a bit to have a better-defined contract.
- Inline the methods from `ConfigFileTypeHandler` into `ConfigTracker`.
- Use the correct mechanism to create default configurations. This suppresses the ugly `Incorrect key xxx was corrected from xxx to its default, xxx.` warnings on the first startup.
- Separate access to registered configs via the new `ModConfigs` from the internal `ConfigTracker`.
- Add a few basic config loading and reloading tests.
- Design the multithreading model for configs:
  - Some operations such as firing events now need to take a lock. This makes sure that reloading and loading cannot fire concurrently.
  - Config objects themselves are threadsafe, and may in fact get reloaded at any time from another thread.
  - If a Config changes, the values will eventually be updated, and an event will eventually be fired to notify the mod.
- Fire ModConfigEvent.Reloading on the client-side when a new config is received.


loader/src/main/java/net/neoforged/fml/config/ConfigTracker.java
```diff
@@ -68,6 +98,33 @@ public void unloadConfigs(ModConfig.Type type) {
         this.configSets.get(type).forEach(this::closeConfig);
     }
 
+    public void openConfig(final ModConfig config, final Path configBasePath, @Nullable Path configOverrideBasePath) {
```
TelepathicGrunt, line 101: I never liked this name. I would see open config as like a fileStream where I read a file contents. Maybe loadConfig would be a better name here? But you got loadConfig already below. hmmm

Technici4n: The idea is that you might load the config multiple times, I suppose. We also have `closeConfig`. I suppose that `open`/`close` is more about the action of opening/closing the `FileConfig`, which might be `loaded` from multiple times over its lifetime?

TelepathicGrunt: So it's more about when the file is being watched or not? `trackConfig` `untrackConfig`

Technici4n: Kind of. It also controls when you can query config values.

TelepathicGrunt: Idk of anything better than `trackConfig` `untrackConfig` but if you want to keep current name, you can resolve this convo

loader/src/main/java/net/neoforged/fml/config/ConfigTracker.java
```diff
@@ -68,6 +98,33 @@ public void unloadConfigs(ModConfig.Type type) {
         this.configSets.get(type).forEach(this::closeConfig);
     }
 
+    public void openConfig(ModConfig config, Path configBasePath, @Nullable Path configOverrideBasePath) {
+        LOGGER.trace(CONFIG, "Loading config file type {} at {} for {}", config.getType(), config.getFileName(), config.getModId());
+        if (config.config != null) {
+            LOGGER.warn("Opening a config that was already loaded with value {} at path {}", config.config, config.getFileName());
```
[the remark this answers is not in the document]
Technici4n: It probably should yeah. Else nothing will close the old file.

loader/src/main/java/net/neoforged/fml/config/ConfigTracker.java
```diff
@@ -79,44 +136,122 @@ private Path resolveBasePath(ModConfig config, Path configBasePath, @Nullable Pa
...
+            }
+        }
+    }
+
+    private static void unload(ModConfig config) {
+        if (FMLConfig.getBoolConfigValue(FMLConfig.ConfigValue.DISABLE_CONFIG_WATCHER))
+            return;
+        Path configPath = config.getFullPath();
+        try {
+            FileWatcher.defaultInstance().removeWatch(configPath);
```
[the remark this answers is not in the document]
Technici4n: That could add a large delay if we are closing 100s of files one by one. This is why I haven't implemented waiting for the watcher yet: we need to submit all the futures and then block on them all at once.

PR #176, opened 2024-07-05
Crash at most once.

shartte:
Early display in general could use some love. But this small change should fix the crashes piling on top of each other.

https://github.com/neoforged/NeoForge/issues/1235


earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/DisplayWindow.java
```diff
@@ -299,13 +298,10 @@ public String getGLVersion() {
...
+    private final ReentrantLock crashLock = new ReentrantLock();
+
     private void crashElegantly(String errorDetails) {
-        String qrText;
-        try (var is = new BufferedReader(new InputStreamReader(getClass().getResourceAsStream("/glfailure.txt")))) {
-            qrText = is.lines().collect(Collectors.joining("\n"));
-        } catch (IOException ioe) {
-            qrText = "";
-        }
+        crashLock.lock(); // Crash at most once!
```
Technici4n, line 304: A bit of a nitpick: Would using a `Semaphore` one permit be a bit more idiomatic? Taking a lock and not releasing it is a bit weird.

Technici4n (approved):

PR #177, opened 2024-07-06
Add the ability to install and run a production client and server as well as moddev for testing FML.

shartte:
This adds the ability to:

- Run `:test:installNeoForgeServer` to install a local copy of the NeoForge server into `build/installs/server` using the installers silent-install feature. This also produces several JVM argfiles that capture the required startup parameters to start it.
- Run `:test:installNeoForgeClient` to install a local copy of the NeoForge client into `build/installs/client` using the installers silent-install feature. This also produces several JVM argfiles that capture the required startup parameters to start it.
- Run `:test:runProductionClient` to start a NeoForge client from the local install using the current projects FML components.
- Run `:test:runProductionServer` to start a NeoForge server from the local install using the current projects FML components.
- Run `:test:runClient` to start a moddev client, using the local FML
- Run `:test:runServer` to start a moddev server, using the local FML


tests/build.gradle
```diff
@@ -0,0 +1,122 @@
...
+        dependencies {
+            classpath files(tasks.named("installNeoForgeClient").map { it.obfuscatedClientJar })
+        }
+    }
+    productionServer {
+        ideName = "Server (Production)"
+        workingDirectory = neoForgeInstallations.server.directory
+        jvmArguments.add(neoForgeInstallations.server.neoForgeJvmArgFile.map { "@" + it })
+        programArguments.add(neoForgeInstallations.server.neoForgeMainClassArgFile.map { "@" + it })
+        programArguments.add(neoForgeInstallations.server.neoForgeProgramArgFile.map { "@" + it })
```
Technici4n, line 74: This task should capture a dependency on `installProductionServer` somehow.

tests/build.gradle
```diff
@@ -0,0 +1,122 @@
...
+        } else {
+            throw new GradleException("Unsupported operating system: " + osName);
+        }
+
+        details.closestMatch(osName);
+    }
+}
+
+var ATTRIBUTE_DISTRIBUTION = Attribute.of("net.neoforged.distribution", String.class);
+var ATTRIBUTE_OPERATING_SYSTEM = Attribute.of("net.neoforged.operatingsystem", String.class);
```
Technici4n, line 107: Should we move all of this stuff to the API of MDG?

buildSrc/src/main/java/fmlbuild/RunConfigurationsPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+            throw new GradleException("Cannot remove installations once they have been registered");
+        });
+
+        project.afterEvaluate(ignored -> {
+            var ijRunConfigs = getIntelliJRunConfigurations(project);
+            if (ijRunConfigs == null) {
+                return;
+            }
+
+            for (var settings : runConfigurations) {
```
Technici4n, line 108: Currently these runs are outright broken anyway. We should maybe add Gradle task runs directly.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.8-bin.zip
```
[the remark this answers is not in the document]
Technici4n: I guess it's fine with `-bin`; most FML developers won't need the full sources.

marchermans:
What is the usecase for this?

You can just includeBuild FML in NeoForge and you get the same result. Is this for production jars or what?

marchermans (changes requested):
I can understand why this is needed.
But I don't think that using NFRT for just the asset download is a great idea.

I would really welcome it if we created an asset download library or something. I have been struggling with this in NG as well, and would welcome it for sure.

Alternatively I think that it should be build in into the installer if that is not wanted, potentially via installer utils.

Does that make sense?

Matyrobbrt:
> But I don't think that using NFRT for just the asset download is a great idea.
Why is that?

test/build.gradle
```diff
@@ -0,0 +1,122 @@
...
+        workingDirectory = neoForgeInstallations.server.directory
+        jvmArgs.add(neoForgeInstallations.server.neoForgeJvmArgFile.map { "@" + it })
+        programArgs.add(neoForgeInstallations.server.neoForgeMainClassArgFile.map { "@" + it })
+        programArgs.add(neoForgeInstallations.server.neoForgeProgramArgFile.map { "@" + it })
+    }
+}
+
+// The production server configuration has to be given the attribute to get server dependencies
+
+abstract class DistributionDisambiguation implements AttributeDisambiguationRule<String> {
```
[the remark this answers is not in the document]
Technici4n: Okay we can leave it here for now.

tests/build.gradle
Technici4n: Note that there was an MDG bug there. I suppose you copied the correct version over. 😉

buildSrc/src/main/java/fmlbuild/InstallProductionClientTask.java
```diff
@@ -0,0 +1,325 @@
...
+        execOperations.javaexec(spec -> {
+            spec.classpath(getNfrt().getSingleFile());
+            spec.args(
+                    "download-assets",
+                    "--minecraft-version",
+                    minecraftVersion,
+                    "--output-properties-to",
+                    assetPropertiesFile.getAbsolutePath()
+            );
+        });
```
marchermans, line 174: Instead of using NFRT I would like to see its asset downloading code in a dedicated library. We have many more locations this could be usefull, for example in the installer, and in neogradle.

Additionally downloading assets is a dedicated unique task that is not dependent on NeoForm

marchermans (changes requested):

buildSrc/src/main/java/fmlbuild/InstallProductionClientTask.java
marchermans: We can merge this for now, but asset downloading should be extracted from NFRT and moved to a library so that it is reuseable.

Technici4n (approved):

PR #178, opened 2024-07-10
Config cleanup, part 2

Technici4n:
Follow-up to #172.

- Stop using `FileConfig`, and instead read/write files manually.
- Pass an `ILoadedConfig` to the `IConfigSpec` so it has a backreference that can be used to save the file and fire events after modification.
- Remove unused config tracking inside the `ModContainer`s. The configs are already accessible via `ModConfigs`.

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/config/IConfigSpec.java
+++ b/loader/src/main/java/net/neoforged/fml/config/IConfigSpec.java
@@ -18,8 +18,7 @@ import org.jetbrains.annotations.Nullable;
  * <h3>Thread-safety</h3>
  * <p>The {@link Config} objects themselves are thread-safe, but mod config code should not be assumed to be thread-safe in general.
  * FML will guard event dispatches behind a lock when necessary.
- * A spec can safely mutate {@code Config} objects, but should let FML fire events,
- * for example using {@link ModConfig#onConfigChanged()}.
+ * A spec can safely mutate {@code Config} objects, but should let FML fire events and saving the file by calling {@link ILoadedConfig#save()}.
  */
 public interface IConfigSpec {
     /**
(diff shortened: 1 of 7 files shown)


loader/src/main/java/net/neoforged/fml/config/IConfigSpec.java
```diff
@@ -53,5 +52,20 @@ public interface IConfigSpec {
...
      * The config is guaranteed to be valid according to {@link #isCorrect}.
      */
-    void acceptConfig(@Nullable CommentedConfig config);
+    void acceptConfig(@Nullable ILoadedConfig config);
+
+    sealed interface ILoadedConfig permits LoadedConfig {
+        /**
+         * Accesses the current config.
+         *
+         * <p>The config locks internally, and is therefore safe to
```
Matyrobbrt, line 61:
```suggestion
         * <p>
         * The config locks internally, and is therefore safe to
```

loader/src/main/java/net/neoforged/fml/config/IConfigSpec.java
```diff
@@ -53,5 +52,20 @@ public interface IConfigSpec {
      * This is called on loading and on reloading.
      * The config is guaranteed to be valid according to {@link #isCorrect}.
      */
-    void acceptConfig(@Nullable CommentedConfig config);
+    void acceptConfig(@Nullable ILoadedConfig config);
+
+    sealed interface ILoadedConfig permits LoadedConfig {
```
Matyrobbrt, line 57: This is a sealed interface, so is there any reason to have it over a normal class?

Matyrobbrt (changes requested):

loader/src/main/java/net/neoforged/fml/config/IConfigSpec.java
Technici4n: Just hiding implementation details. :wink:

loader/src/main/java/net/neoforged/fml/config/IConfigSpec.java
Technici4n: No :P

Matyrobbrt (approved):

PR #181, opened 2024-07-13
Fix NetworkedEnum annotation retention

XFactHD:
This PR fixes the retention of the `NetworkCheck` annotation to allow the annotation to be read reflectively by `ExtensionInfo.nonExtended()`.


XFactHD:
Nope, none of the tests currently touches the extension info getter. I'll add something in a bit.

PR #184, opened 2024-07-15
Add a method to validate the config spec

Matyrobbrt:
Add a `IConfigSpec#validateSpec` method to validate the specification in the context of a mod config, i.e. to check if a value restart type is valid for a given config type.

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/config/ConfigTracker.java
+++ b/loader/src/main/java/net/neoforged/fml/config/ConfigTracker.java
@@ -87,6 +87,8 @@ public class ConfigTracker {
     public ModConfig registerConfig(ModConfig.Type type, IConfigSpec spec, ModContainer container, String fileName) {
         var lock = locksByMod.computeIfAbsent(container.getModId(), m -> new ReentrantLock());
         var modConfig = new ModConfig(type, spec, container, fileName, lock);
+        spec.validateSpec(modConfig);
+
         trackConfig(modConfig);
 
         if (modConfig.getType() == ModConfig.Type.STARTUP) {
--- a/loader/src/main/java/net/neoforged/fml/config/IConfigSpec.java
+++ b/loader/src/main/java/net/neoforged/fml/config/IConfigSpec.java
@@ -27,6 +27,13 @@ public interface IConfigSpec {
     boolean isEmpty();
 
     /**
+     * Validate this specification in the context of the given {@code config}.
+     *
+     * @param config the configuration this spec is used by
+     */
+    void validateSpec(ModConfig config);
+
+    /**
      * Checks that a config is correct.
      * If this function returns {@code false},
      * a backup is made (except for initial creation) then the config is fed through {@link #correct}.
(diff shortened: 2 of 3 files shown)


marchermans:
What is the context of this change?
Why do you want to add this?

Matyrobbrt:
This PR was made specifically for https://github.com/neoforged/NeoForge/pull/1284 in order to validate if a restart type is compatible with a config type (i.e. `GAME` restart not being compatible with `SERVRE` configs)

marchermans:
The mentioned PR, does not update FML, nor does it use the newly introduced method `validateSpec`.

Could you elaborate the idea for this change?

Matyrobbrt:
I'm not sure what I should elaborate more on. It's quite simple, make sure that the specification is valid when used by a specific config type. So that the neo config spec can throw when someone attempts to use a game restart flag for a server config (which doesn't make sense)

Matyrobbrt:
> Okey..... But this is not used in that PR
Yet, I just made this PR yesterday...

> Also I think a GameRestart flag on Server config makes a lot of sense. Why would we disallow that. It indicates that the server needs to be restarted.
That's a world restart which is a separate flag that actually works with SP worlds.

marchermans:
> > Okey..... But this is not used in that PR
> 
> Yet, I just made this PR yesterday...
> 
> > Also I think a GameRestart flag on Server config makes a lot of sense. Why would we disallow that. It indicates that the server needs to be restarted.
> 
> That's a world restart which is a separate flag that actually works with SP worlds.

Isn't a game restart and a server restart on the server the same? So why would seperating this out matter?

TelepathicGrunt:
> Isn't a game restart and a server restart on the server the same? So why would separating this out matter?

I would just treat both flags the same on a server config tbh instead of erroring for one. Erroring will lead to modder confusion as game restart sounds like it should restart server.

Matyrobbrt:
The problem is that the flags ought to behave the same in all contexts, and they are radically different in a non singleton context (singleplayer). These flags are, if anything, more user facing than modder facing and their intent and effect need to be clear to users.

marchermans:
> The problem is that the flags ought to behave the same in all contexts, and they are radically different in a non singleton context (singleplayer). These flags are, if anything, more user facing than modder facing and their intent and effect need to be clear to users.
> 
> While for a physical server, game and server restart are the same, for a single player world they're completely different concepts.

In what way are they completely different contexts?
Both indicate exactly what they are supposed to do, whether in dedicated server, or in integrated server. Just in dedicated server the only way to restart a world is by restarting the game. So I agree with grunt here. I see no reason to differentiate on these flags, and as such no reason to merge this PR.

TelepathicGrunt:
Change my mind. Blocking game restart for server config to require world restart makes more sense in single player context and not cause users to restart the game when they don’t need to. Just exit and re-enter world

marchermans:
> Change my mind. Blocking game restart for server config to require world restart makes more sense in single player context and not cause users to restart the game when they don’t need to. Just exit and re-enter world

Agreed, but again that discussion is now not on the record here.
So it will get lost to time.

However approving.

Technici4n (approved):

PR #185, opened 2024-07-18
Add simple getter for configs by mod and the loaded configs

Technici4n:
Quite simple and useful, especially for alternative config screen implementations that have a different strategy. Requested by @Speiger in https://github.com/neoforged/NeoForge/issues/1336.

Fixes https://github.com/neoforged/NeoForge/issues/1336.


Speiger:
@Technici4n why List.copyOf instead of Collections.unmodifiableList? (Wrap instead of copy i mean)

TelepathicGrunt (approved):

Issue #98, opened 2024-03-17
 Make earlydisplay work on Wayland if possible.


heipiao233:
Wayland is becoming more and more popular between Linux desktop users. Many desktop environments support (or plan to support) Wayland. GLFW and LWJGL supports Wayland, too.

PR #187, opened 2024-07-22
Wayland early display window

wired-tomato:
Wayland early display window implementation, requested in #98


wired-tomato:
the window will load up with wayland on GLFW 3.3+, the icon won't load unless GLFW version 3.4+ is in use, everything else should work fine

wired-tomato:
notably there is a GLFW error printed to the console `65548: Wayland: The platform does not support setting the cursor position` when opening inventories & some other screens, however it does not cause any crashes or actual issues and can be solved with a mixin as demonstrated in my fabric wayland compat mod [WayGL](https://github.com/wired-tomato/WayGL)

theendercore:
Soooo.. 
How's the testing going :3

DerCommander323:
> > `glfwWindowHint(GLFW_FOCUS_ON_SHOW, GLFW_FALSE);`
> 
> Why is this actually necessary for Wayland support?

After some testing, it seems that currently KDE Plasma incorrectly reports the Window as "not responding" on initial focus. After alt tabbing away and back onto the window this goes away. I assume that's what this hint is for.

wired-tomato:
I had initially removed the hint after testing for that issue on my own system, I've added it back to see if it fixes the issue.

PR #189, opened 2024-08-01
Support ordering of dependency locators

Su5eD:
Adds support for ordering dependency locators. This is useful for mods that rely on information about previously located mods in the scanning process.


Su5eD:
Yes

Matyrobbrt (approved):

PR #204, opened 2024-09-16
Fix window icon not being set on startup.

shartte:
Fixes the NeoForge icon not correctly being set since the `put` advanced the buffer thus making LWJGL think the icon array is empty.

image

image

p.s.: the icon will switch back to a Minecraft icon as normal when Vanilla takes over.


sciwhiz12 (approved):
Seems to work.

embeddedt (approved):

PR #206, opened 2024-10-23
Fix: Disambiguate libraries by their module name.

LlamaLad7:
MixinExtras is a library rather than a mod, at the request of the team, but this means that JiJ'ed versions are not correctly grouped since the current logic uses the filename.
This allows users to JiJ new stable versions of mixinextras.
There is another bug with the module name detection for beta versions, but I will fix that separately by setting an `Automatic-Module-Name`.

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/locators/JarInJarDependencyLocator.java
+++ b/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/locators/JarInJarDependencyLocator.java
@@ -109,8 +109,13 @@ public class JarInJarDependencyLocator implements IDependencyLocator {
     }
 
     protected String identifyMod(final IModFile modFile) {
-        if (modFile.getModFileInfo() == null || modFile.getModInfos().isEmpty()) {
+        if (modFile.getModFileInfo() == null) {
             return modFile.getFileName();
+        }
+        // If this is a library, it won't have any mod IDs, so we use the module name instead.
+        if (modFile.getModInfos().isEmpty()) {
+            // Prefix to ensure this cannot collide with any true mod ID.
+            return "library:" + modFile.getModFileInfo().moduleName();
         }
 
         return modFile.getModInfos().stream().map(IModInfo::getModId).collect(Collectors.joining());
--- a/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/readers/JarModsDotTomlModFileReader.java
+++ b/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/readers/JarModsDotTomlModFileReader.java
@@ -140,12 +140,12 @@ public class JarModsDotTomlModFileReader implements IModFileReader {
             return configurable;
         }
 
-        // These Should never be called as it's only called from ModJarMetadata.version and we bypass that
         @Override
         public String moduleName() {
             return mod.getSecureJar().name();
         }
 
+        // These Should never be called as it's only called from ModJarMetadata.version and we bypass that
         @Override
         public String versionString() {
             return null;


loader/src/main/java/net/neoforged/fml/loading/moddiscovery/locators/JarInJarDependencyLocator.java
```diff
@@ -109,9 +109,12 @@ private String formatError(final ModWithVersionRange modWithVersionRange) {
     }
 
     protected String identifyMod(final IModFile modFile) {
-        if (modFile.getModFileInfo() == null || modFile.getModInfos().isEmpty()) {
+        if (modFile.getModFileInfo() == null) {
             return modFile.getFileName();
         }
+        if (modFile.getModInfos().isEmpty()) {
```
Technici4n, line 115: This could use a comment explaining it's for libraries. It would also be good to explain why the `modFile.getModFileInfo() == null` check above is necessary but that's not related to your PR.

LlamaLad7:
One question is what should happen when a modid and a module name overlap? Should we add some kind of `library:` prefix so that can't happen?

Technici4n:
Maybe we can do that yeah, just in case.

LlamaLad7:
Ok all done

marchermans:
> MixinExtras is a library rather than a mod, at the request of the team, but this means that JiJ'ed versions are not correctly grouped since the current logic uses the filename.

That is not true, grouping and selection of the jars, happens solely via the metadata that is included in the json of the containing jar, the identifier information is not actually used for this.

What is the current exact problem?

marchermans:
Selection happens solely via this select method: <https://github.com/neoforged/JarJar/blob/b9338af0d631251d5ed6779b14c66bd419a2cf37/selector/src/main/java/net/neoforged/jarjar/selection/JarSelector.java#L181-L218> that directly selects via the `ContainedJarMetadata` which is created directly from the metadata of the containing jar.

loader/src/main/java/net/neoforged/fml/loading/moddiscovery/locators/JarInJarDependencyLocator.java
```diff
@@ -109,9 +109,14 @@ private String formatError(final ModWithVersionRange modWithVersionRange) {
...
+        if (modFile.getModFileInfo() == null) {
             return modFile.getFileName();
         }
+        // If this is a library, it won't have any mod IDs, so we use the module name instead.
+        if (modFile.getModInfos().isEmpty()) {
+            // Prefix to ensure this cannot collide with any true mod ID.
+            return "library:" + modFile.getModFileInfo().moduleName();
+        }
 
         return modFile.getModInfos().stream().map(IModInfo::getModId).collect(Collectors.joining());
```
lukebemish, line 121: If we're already fixing up this logic, can we fix this logic here too, by joining them with something that _can't_ be in a mod ID, like `+` or `,`? The current logic would treat a jar containing the mods `foo` and `bar` as identical to a jar containing the mod `foobar`, which it is not.

lukebemish (changes requested):

loader/src/main/java/net/neoforged/fml/loading/moddiscovery/locators/JarInJarDependencyLocator.java
LlamaLad7: out of scope IMO

lukebemish: The other option is to take just the _first_ mod ID -- which may be more sensible, as that is what becomes the module name. That said -- is there a reason we don't just use the module name in _general_ here? After all, it'll be unique for different mods and for other stuff, but the same between multiple mods that are the "same" mod or whatever.

Technici4n (approved):
Going to merge this, looks fine for now.

PR #210, opened 2024-10-25
Remove "maximum GL" window feature & add debug context

IMS212:
This adds a debug context config option, and removes the maximum GL spam.

The GL version is calculated from `GLCapabilities` now.


IMS212:
Requires #245 before rebase.

Issue #201, opened 2024-09-03
Annotation for loading class if mod is present


CodexAdrian:
Its common practice for mods to add apis or features other devs can use to add integration. For optional integration its common place for a dev to put the compat in a different class and then call that class with an if check for if the mod is loaded. This is fine, but I think itd be super helpful to have something to conditionally load a class with an annotation. like
```
@Mod(value = "modid", required = "argonauts")
```
So that the class will only ever load if the mod "argonauts" is present.

Technici4n:
Why not just load the class via reflection in the mod, instead of adding more magic just to save a few LOC?

PR #212, opened 2024-11-12
Conditional mods & event bus subscribers

wired-tomato:
Conditional class loading described in #201 

```java
@Mod(
        value = "testmod",
        dependencies = { 
                @ChainDependency(value = @Dependency({ "neoforge" }), operator = Operator.AND),
                @ChainDependency(value = @Dependency(value = { "othermod", "othermod2" }, condition = Condition.NONE_PRESENT))
        }
)
public class TestMod {
    public TestMod() {
        //called if neoforge is present and othermod and othermod2 are both not present
    }
}
```

```java
@EventBusSubscriber(
        modid = "testmod",
        dependencies = {
                @ChainDependency(value = @Dependency({ "neoforge" }), operator = Operator.AND),
                @ChainDependency(value = @Dependency(value = { "othermod", "othermod2" }, condition = Condition.NONE_PRESENT))
        }
)
public class TestModEvents {
    //Subscribed if neoforge is present and othermod and othermod2 are not present
    public void configEvent(ModConfigEvent event) {}
}
```

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/common/ChainDependency.java
+++ b/loader/src/main/java/net/neoforged/fml/common/ChainDependency.java
@@ -1,0 +1,16 @@
+/*
+ * Copyright (c) NeoForged and contributors
+ * SPDX-License-Identifier: LGPL-2.1-only
+ */
+
+package net.neoforged.fml.common;
+
+public @interface ChainDependency {
+    Dependency value();
+
+    Operator operator() default Operator.AND;
+
+    enum Operator {
+        AND, OR
+    }
+}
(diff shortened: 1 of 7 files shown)


tmvkrpxl0:
Wait who is that guy

Gaming32:
Looking at the beginning of the impl, it seems you can use `dependencies = @ChainDependency(@Dependency("modid"))`. Though that's still too verbose in my opinion.

Technici4n:
I think that OR conditions are quite suspicious if the goal is to avoid crashes because of missing classes.

Technici4n:
The main goal of checking for dependencies in the annotation like this, is to prevent accidentally loading code that reference nonexisting classes. If you add NOT or OR dependencies, that flies out of the window, and you might as well just add the following code instead of writing convoluted annotations:
```java
if (ModList.get().isModLoaded("a") && !ModList.get().isModLoaded("b")) {
    bus.register(AButNotBEventHandler.class);
}
```

wired-tomato:
Removed ORs and NOTs, `@DependsOn` now only takes an array of modids that must all be present for the class to be loaded

PR #214, opened 2024-11-14
Add dependency overrides

### LINKED ISSUE #52: Add dependency override for player/pack maker to bypass limits set in mod's mods.toml file
With the upcoming `incompatible` and `conflicts` clauses being added to mods.toml, it reminded me of how Forge ecosystem has kinda a problem with the mods.toml file limits. Specifically around what user/pack makers can control.

I am in the camp that the user/pack maker should have final say on what MC version and what mods a mod can run with. So there should be a way to have users/pack makers be able to override `version ranges`, `incompatible`, and `conflicts` for any given mod in their pack. Fabric has dependency overrides for a long while and I have seen users/pack makers make use of it to keep their packs running and updated properly. There has been no influx of reports to modders when their mods are forced to load so any concerns about issue report influx is not really a thing.
https://fabricmc.net/wiki/tutorial:dependency_overrides

Consider this, a modder does not maintain their mods forever and may not exist in modding scene forever. Uers and pack makers will continue to make packs and play on any Minecraft version. With the current system, if a mod does not load in a pack due to a mods.toml restriction, users/pack makers are stuck and cannot do anything forever.

------------

Example 1:

Mod A and Mod B together will crash Mod A. Mod A as a result sets Mod B as `incompatible` to stop loading with it. Mod A's dev eventually leaves the modding scene. Mob B releases an update that fixes the crash. However, users and pack makers still cannot use the mods together because of the `incompatible` clause being unable to be overridden. This can happen with the version range too as mods can set "" as the range to not run with any version of another mod. 

Solution 1: Fork the mod, change the mods.toml file, reupload it to CurseForge/Modrinth, and then add it to modpack to distribute. Problem of that? Ridiculously excessive, taking points and downloads away from original dev, requires mod to be open source and buildable, pack makers to know how to build project, and for CurseForge/Modrinth to allow a reuploaded clone to exist with only a mods.toml change. Awful in every sense of the word.

Solution 2: Allow pack makers to specify a dependency override for the mod in their modpack. Clean, simple, and works.

-----------

Example 2:

Mod A's dev is a jerk and decides to mark Mod B as `incompatible` or `version range` lock out Mod B in latest Minecraft version. Pack makers now have to either remove Mob B to update Mod A to fix crashes or whatever or remove Mod A. this is a blatant misuse of the mods.toml dependencies and users/pack makers have no way of bypassing it cleanly without yeeting mods or forking and reuploading. Dependency Overrides allow users/pack makers to say "no" and bypass the mods.toml misuse to keep the pack updated properly and running.

Matyrobbrt:
This implements a simple system that allows users to override dependency constraints using the fml config by adding an entry as follows:
```toml
dependencyOverrides.target = ["-dep1", "+dep2"]
```
Minus can be used to soften dependency constraints (version range becomes unbounded, type becomes optional, but the ordering stays the same) and plus can be used to add an ordering dependency (the target mod will load AFTER the dependency) 
All overrides are marked in the log at WARN level for visibility. 
Closes #52


TelepathicGrunt:
“dependencyOverries”
Typo in description btw. Just double check typo isn’t in the code too.

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
```diff
@@ -81,12 +86,14 @@ private static Object maxThreads(final Object value) {
 
     private static final Logger LOGGER = LogUtils.getLogger();
     private static final FMLConfig INSTANCE = new FMLConfig();
+    private static final Map<String, List<DependencyOverride>> DEPENDENCY_OVERRIDES = new HashMap<>();
```
Technici4n, line 89: Nitpick: this is a mutable field so it should use lower camel case, even if it's constant. :P

Matyrobbrt: The field is not mutable, it's final.

Technici4n: It is final, but it contains a mutable value. Just look at the lines below :P

loader/src/main/java/net/neoforged/fml/loading/ModSorter.java
```diff
@@ -120,6 +120,14 @@ private void sort() {
                 .map(IModInfo::getDependencies).<IModInfo.ModVersion>mapMulti(Iterable::forEach)
                 .forEach(dep -> addDependency(graph, dep));
 
+        FMLConfig.getDependencyOverrides().forEach((id, overrides) -> {
+            for (FMLConfig.DependencyOverride override : overrides) {
+                if (!override.remove()) {
+                    graph.putEdge((ModInfo) modIdNameLookup.get(override.modId()), (ModInfo) modIdNameLookup.get(id));
```
Technici4n, line 126: What if the mods don't exist? We add null to the graph? :P

loader/src/main/resources/META-INF/defaultfmlconfig.toml
```diff
@@ -19,3 +19,8 @@ earlyWindowControl = true
 #Max threads for early initialization parallelism,  -1 is based on processor count
 maxThreads = -1
 
+# Define dependency overrides below
+# Dependency overrides can be used to forcibly remove a dependency constraint from a mod or to force a mod to load AFTER another mod
+# Using dependency overrides can cause issues. Use at your own risk.
+# Example dependency override for the mod with the id 'targetMod': dependency constraints (incompatibility clauses or restrictive version ranges) against mod 'dep1' are removed, and the mod will now load after the mod 'dep2'
+# dependencyOverrides.targetMod = ["-dep1", "+dep2"]
```
Technici4n, line 26: We might want to show the expanded syntax instead? idk

```toml
[dependencyOverrides]
targetMod = ["-dep1", "+dep2"]
targetMod2 = ["-dep3", "-dep4"]
```

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
```diff
@@ -119,6 +126,32 @@ public static void load() {
...
+                var asList = object instanceof List<?> ls ? ls : List.of(object);
+                var overrides = DEPENDENCY_OVERRIDES.computeIfAbsent(modId, k -> new ArrayList<>());
+                for (Object o : asList) {
+                    var str = (String) o;
+                    var start = str.charAt(0);
+                    if (start != '+' && start != '-') {
+                        LOGGER.error("Found invalid dependency override for mod '{}'. Expected +/- in override '{}'. Did you forget to specify the override type?", modId, str);
+                    } else {
+                        var removal = start == '-';
+                        var depMod = str.substring(1);
```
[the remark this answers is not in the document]
Matyrobbrt: We can't check it at this time no, it will have to be checked where it's used.

loader/src/main/java/net/neoforged/fml/loading/ModSorter.java
Matyrobbrt: We will need a check yes, though I'm not sure if it should be a crash or a warning.

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
```diff
@@ -81,12 +89,17 @@ private static Object maxThreads(final Object value) {
 
     private static final Logger LOGGER = LogUtils.getLogger();
     private static final FMLConfig INSTANCE = new FMLConfig();
+    private static Map<String, List<DependencyOverride>> dependencyOverrides;
```
Technici4n, line 92: Add `@Nullable` or initialize with an empty map.

loader/src/main/resources/META-INF/defaultfmlconfig.toml
Technici4n, line 26: Unfortunately this file is not actually used. When I delete the `config/fml.toml`, it is NightConfig's correction system that is used. We should document this in the generated config file somehow.

Technici4n (changes requested):

loader/src/main/resources/META-INF/defaultfmlconfig.toml
Matyrobbrt: Ugh, you're right, and night configs have no way of storing root comments...

Matyrobbrt: I'll add an empty `dependencyOverrides = {}`

Matyrobbrt: How are you testing?

Technici4n: Okay nevermind, it was hidden in the middle of the file. 😓
But it should be at the end, why is it not...

Matyrobbrt: I have no idea what ordering fml uses

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
```diff
@@ -103,6 +112,17 @@ private static Object maxThreads(final Object value) {
         for (ConfigValue cv : ConfigValue.values()) {
             cv.buildConfigEntry(configSpec, configComments);
         }
+
+        // Make sure that we don't end up "correcting" the config and removing dependency overrides
+        // Since we're not writing them by default, the default value can be null and we accept any objects (parsing and validation is done when the config is loaded)
```
Technici4n, line 117: We do write them by default now so this comment should be updated.

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
```diff
@@ -103,6 +112,17 @@ private static Object maxThreads(final Object value) {
...
+        configSpec.define("dependencyOverrides", Map::of, object -> true);
```
Technici4n, line 118: The `Map::of` is actually useless and should be replaced again by `() -> null`. Sorry 😅

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
```diff
@@ -103,6 +112,17 @@ private static Object maxThreads(final Object value) {
...
+        configComments.set("dependencyOverrides", configComments.createSubConfig());
+        configComments.setComment("dependencyOverrides", """
```
Technici4n, line 120: To actually write this out you need to add the following lines down where the default config is written: (i.e. after `configSpec.correct(this.configData);`)
```diff
+            // Since dependency overrides have an empty validator, they need to be added manually.
+            // (Correct doesn't correct an absent value since it's valid).
+            this.configData.set("dependencyOverrides", this.configData.createSubConfig());
```

Technici4n (changes requested):

PR #218, opened 2024-11-24
Clean up default config writing

Technici4n:
- Get rid of the `CommentedFileConfig` in favor of explicitly reading and writing a config.
- Always use a `LinkedHashMap` as the backend of a config when it affects ordering.
- Get rid of `defaultfmlconfig.toml`, and replace it by just correcting an empty config.


loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
```diff
@@ -64,11 +69,20 @@ void buildConfigEntry(ConfigSpec spec, CommentedConfig commentedConfig) {
...
         @SuppressWarnings("unchecked")
-        private <T> T getConfigValue(CommentedFileConfig config) {
+        private <T> T getConfigValue(CommentedConfig config) {
             return (T) this.entryFunction.apply(config != null ? config.get(this.entry) : this.defaultValue);
         }
 
+        /**
+         * @deprecated Use {@link FMLConfig#updateConfig(ConfigValue, Object)} instead.
+         */
+        @Deprecated(forRemoval = true)
```
Matyrobbrt, line 79: We can remove it now

Matyrobbrt (changes requested):

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
Technici4n: We can but you'll have to reintroduce it when backporting to 1.21.1. I thought we might as well wait. 😅

Matyrobbrt: Ah right, fair enough.

Matyrobbrt (approved):

PR #237, opened 2025-01-18
Make mixin entries from the mods.toml optional.

marchermans:
This allows modders to specify that a mixin config is optional.
It will not be loaded if it does not exist.

Usecase:
- Compatibility mixins loaded from a seperate sourceset, for example when creating compatibility layers with shader mods.

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ModFileParser.java
+++ b/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ModFileParser.java
@@ -16,12 +16,11 @@ import java.io.IOException;
 import java.lang.reflect.Type;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
+
 import net.neoforged.fml.loading.LogMarkers;
 import net.neoforged.fml.loading.moddiscovery.readers.JarModsDotTomlModFileReader;
+import net.neoforged.neoforgespi.language.IConfigurable;
 import net.neoforged.neoforgespi.language.IModFileInfo;
 import net.neoforged.neoforgespi.locating.IModFile;
 import net.neoforged.neoforgespi.locating.InvalidModFileException;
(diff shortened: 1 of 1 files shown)


marchermans:
> Interesting, but what is the advantage over the current convention of overriding `shouldApplyMixin` in a mixin plugin (which is also more flexible)?

Mixin crashes before that if the json is not found in the jar.
So when I am running in none shader compat mode, it just hard crashes as the mixin json it self lives in the other sourceset with the compat code.

loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ModFileParser.java
```diff
@@ -85,17 +88,32 @@ protected static List<CoreModFile> getCoreMods(final ModFile modFile) {
                 .toList();
     }
 
+    private record PotentialMixin(String name, boolean optional) {}
```
FiniteReality, line 91: Suggestion: `PotentialMixinConfig` since it's a mixin config we're specifying, not an individual mixin

marchermans: Right, let me fix that.

Technici4n:
That's really weird because it is a dev-only feature. I wonder if adding mixin configs that depend on some mods being loaded would not be a more general solution. :thinking:

marchermans:
> That's really weird because it is a dev-only feature. I wonder if adding mixin configs that depend on some mods being loaded would not be a more general solution. 🤔

Maybe, but this is a good first step, and when the need arises we can definitely iterate on this design and expand the conditions.

Technici4n:
It's weird to ever have this in a production environment. And in dev there should be a suitable way to trim the unwanted mixin entries, no? (E.g. Gradle magic)

We should also think about why this isn't done with accesstransformers.

marchermans:
> It's weird to ever have this in a production environment. And in dev there should be a suitable way to trim the unwanted mixin entries, no? (E.g. Gradle magic)
> 
> We should also think about why this isn't done with accesstransformers.

It is possible to do with a lot of gradle magic, where gradle edits the file, and injects stuff into it, all in all, with running with IDEA support in the back of my mind (as that is now the default way of running my mods), it is a lot of work, for something I think we should start working towards (aka optional, conditional on mods, etc). This PR provides initial infrastructure to do that (with the capturing record, and filtering logic) to eventually get to full conditional support for this. But we can iterate and see what the community wants or needs.

loader/src/main/java/net/neoforged/fml/loading/LoadingModList.java
```diff
@@ -77,7 +77,13 @@ public void addMixinConfigs() {
                 .map(ModFileInfo::getFile)
                 .forEach(file -> {
                     final String modId = file.getModInfos().get(0).getModId();
-                    file.getMixinConfigs().forEach(cfg -> DeferredMixinConfigRegistration.addMixinConfig(cfg, modId));
+                    file.getMixinConfigs()
+                            .stream()
+                            .filter(potential -> potential.requiredModIds().isEmpty() ||
```
Technici4n, line 82: You can remove the `isEmpty` check. :wink:

loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ModFileParser.java
```diff
@@ -85,17 +87,28 @@ protected static List<CoreModFile> getCoreMods(final ModFile modFile) {
...
     }
 
-    protected static List<String> getMixinConfigs(IModFileInfo modFileInfo) {
+    /**
+     * Represents a potential mixin configuration.
+     *
+     * @param name The name of the mixin configuration.
+     * @param requiredModIds The mod ids that are required for this mixin configuration to be loaded. If empty, will be loaded regardless.
+     */
+    public record PotentialMixinConfig(String name, List<String> requiredModIds) {}
```
Technici4n, line 96: Since this is API, I would probably call this differently. It really is just a Java object model for how a mod declares a mixin config. Maybe `MixinsEntry` or whatever.

Technici4n, line 96: The name field should be called `config` to match the TOML format.

Technici4n, line 96: The mod IDs field should be called `requiredMods` to match the TOML format.

Technici4n (changes requested):

loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ModFileParser.java
marchermans: Given that it is not API (else JCC would not allow this), the naming does not matter, but MixinsEntry is fine to me.

marchermans: Actually I named it `MixinConfig` as the method is named: `getMixinConfigs`

Technici4n (approved):

PR #243, opened 2025-02-13
Increase early window creation timeout to 30 seconds

jellysquid3:
A regression was introduced in Intel Arc driver update 32.0.101.6559 which is causing OpenGL context initialization to take an extremely long time (~12 seconds) when starting Minecraft. This problem affects Vanilla as well, but only NeoForge has a window creation timeout, which results in the game being prematurely terminated.

As a short-term solution, this patch increases the timeout to 30 seconds, which is a long enough period for the driver to take care of what it needs to.

Note that the underlying problem here needs to be reported to Intel, as context creation shouldn't take this long. But seeing as they've rolled out this WHQL driver update, it is the unfortunate situation that we have to deal with it.


Technici4n (approved):

Matyrobbrt:
@neoforged/bots backport 1.21.1

PR #253, opened 2025-03-01
Catch any exception during EBS injection

Matyrobbrt:
For some reason EBS injection only caught and rethrew with more details CNFEs, this PR widens it to all exceptions (like exceptions caused by the dist cleaner)


Matyrobbrt:
@neoforged/bots backport 1.21.1

PR #257, opened 2025-03-02
Add a config to enable debug output and object labels for OpenGL in erly display.

shartte:
Adds an FML option to:
- Enable setting GL debug labels on objects created by early display
- Add a debug group for updating the earlydisplay framebuffer
- Enable the GL debug callback during early display (and make it synchronous) to enable debugging of early display

### DIFF
--- a/earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/GlDebug.java
+++ b/earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/GlDebug.java
@@ -39,7 +39,7 @@ final class GlDebug {
     private static int maxLabelLength;
 
     public static void setCapabilities(GLCapabilities capabilities) {
-        if (!FMLConfig.getBoolConfigValue(FMLConfig.ConfigValue.EARLY_WINDOW_GLDEBUG)) {
+        if (!FMLConfig.getBoolConfigValue(FMLConfig.ConfigValue.DEBUG_OPENGL)) {
             return;
         }
 


earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/GlDebug.java
```diff
@@ -0,0 +1,173 @@
...
+            labelMode = LabelMode.CORE;
+            groupMode = GroupMode.CORE;
+            maxLabelLength = GL32C.glGetInteger(KHRDebug.GL_MAX_LABEL_LENGTH);
+        } else {
+            if (capabilities.GL_EXT_debug_label) {
+                labelMode = LabelMode.EXTENSION;
+                maxLabelLength = 256;
+            }
+            if (capabilities.GL_EXT_debug_marker) {
+                groupMode = GroupMode.EXTENSION;
```
XFactHD, line 57: Is there any case where `GL_EXT_debug_marker` might be true while `GL_EXT_debug_label` is false, causing `maxLabelLength` to not be set to a non-zero value?

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/SimpleBufferBuilder.java
```diff
@@ -390,6 +392,8 @@ public void draw() {
             // but only once, the VAO saves this state
             GlState.bindVertexArray(vao);
             GlState.bindArrayBuffer(vbo);
+            GlDebug.labelVertexArray(vao, label);
+            GlDebug.labelBuffer(vao, label);
```
XFactHD, line 396: Shouldn't this label the `VBO` instead of the `VAO`?

loader/src/main/java/net/neoforged/fml/loading/FMLConfig.java
```diff
@@ -41,6 +41,7 @@ public enum ConfigValue {
         DEFAULT_CONFIG_PATH("defaultConfigPath", "defaultconfigs", "Default config path for servers"),
         DISABLE_OPTIMIZED_DFU("disableOptimizedDFU", Boolean.TRUE, "Disables Optimized DFU client-side - already disabled on servers"),
         EARLY_WINDOW_PROVIDER("earlyWindowProvider", "fmlearlywindow", "Early window provider"),
+        EARLY_WINDOW_GLDEBUG("earlyWindowGlDebug", Boolean.FALSE, "Enable OpenGL debugging in early window provider"),
```
XFactHD, line 44: I think it would be a good idea to make this setting more generic and wire it into `RenderSystem.initRenderer()` in NF (for sync debug output the config would be passed directly to `initRenderer()` and for debug labels the config would be ORed together with vanilla's `renderDebugLabels` command line option). That way GL debugging controlled by this setting would be consistent across the entire game instead of some parts being enabled in one place but not another and, specifically for sync debug output, would also allow this setting to impact the game even if the early window is disabled.

XFactHD (changes requested):

XFactHD (approved):

tmvkrpxl0:
* early

PR #260, opened 2025-03-23
 Rework the interface between NeoForge and the FML early loading screen 

shartte:
The primary goal if this PR is to greatly simplify the interface between NeoForge and FML when it comes to the early loading screen (a.k.a. early display). It removes all direct access from FML to NeoForge (i.e. no more module reads, reflective access to NoVizFallback or NeoForgeLoadingOverlay).
Instead, it introduces an abstraction with just the methods that NeoForge would call on ImmediateWindowHandler, and offers this *nullable* to NeoForge. Nullable is key here since that will allow NeoForge to know whether it has to just fall back to Vanilla methods, which previously were routed through FML using NoVizFallback. This is now gone.

Note that this also removes the OpenGL feature flag since it was previously set using a very awkward bi-directional call between FML and NF, since FML has no access to the actual window or GL context when not using earlydisplay.
If the feature flag is still desired, it must be registered in the ClientModLoader in NeoForge instead, and a fallback value may be provided for server or datagen.


Technici4n:
Can we see what the changes look like on the NeoForge side?

marchermans (approved):

PR #267, opened 2025-03-27
Remove deprecated functionality

shartte:


loader/src/main/java/net/neoforged/fml/InterModComms.java
```diff
@@ -17,49 +17,7 @@
...
-         * @return A {@link Supplier} of the message.
-         *         Use {@link #messageSupplier()}
-         */
-        @SuppressWarnings("unchecked")
-        @Deprecated
-        public final <T> Supplier<T> getMessageSupplier() {
-            return (Supplier<T>) this.messageSupplier;
-        }
-    }
+    public record IMCMessage(String senderModId, String modId, String method, Supplier<?> messageSupplier) {}
```
Technici4n, line 20: Might be nice to keep the javadoc? It's rather trivial, but there are interesting points such as `This is supplied by the caller, or by the active mod container context. Consider it unreliable.`

Technici4n (approved):

PR #271, opened 2025-04-07
Fix: no modid is logged when "Failed to validate feature bounds"

Viola-Siemens:
Currently, FML can not output which mod causes `Failed to validate feature bounds`. The only logged information is the address of the mod info:

image

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/ModLoader.java
+++ b/loader/src/main/java/net/neoforged/fml/ModLoader.java
@@ -114,7 +114,7 @@ public final class ModLoader {
                 .toList();
 
         if (!failedBounds.isEmpty()) {
-            LOGGER.fatal(CORE, "Failed to validate feature bounds for mods: {}", failedBounds);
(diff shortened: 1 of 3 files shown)


Viola-Siemens:
Btw I [cannot run the workflow in my repo](https://github.com/Viola-Siemens/FancyModLoader-forked/actions/runs/14312211498/job/40109769173) so I have to draft the PR first to check if the code is working.

Viola-Siemens:
> Thanks for your PR, but the logging should be based on the mod-loading issues that are created a few lines below. They have a suitable `toString()` method meaning they can just be used as logging arguments as-is.

Thanks for explaining. The key problem is that `modInfo` is not serializable. And also I don't know whether any other places use `toString()` method for other uses and modifying `toString()` method could cause trouble or not. If it's fine, I'll change `visualizedInfo` into `toString` and remove changes from `ModLoader` and `Bound`.

Additionally, I think this change is necessary because the fml crash report may sometimes be replaced by client crash report when some special mods are installed, which makes it harder to debug.

PR #274, opened 2025-04-14
Increase log verbosity of refused event dispatch messages

FiniteReality:
Usually, when a modder has an error in their code, we will spam out a bunch of these errors because the mod state is broken.

However, if it were a true error, you would expect all of our dispatches to be guarded with something to the effect of `if (!bus.isInErrorState()) { bus.dispatchEvent(...); }` which is not the case.

Because of this, it's noise at best and should be warn at highest. Instead, I chose trace because it's TYPICALLY only of use to Neo devs, or more involved debugging from modders.

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/ModLoader.java
+++ b/loader/src/main/java/net/neoforged/fml/ModLoader.java
@@ -330,7 +330,7 @@ public final class ModLoader {
 
     public static <T extends Event & IModBusEvent> void runEventGenerator(Function<ModContainer, T> generator) {
         if (hasErrors()) {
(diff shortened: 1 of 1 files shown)


Technici4n:
Not sure what this really achieves, the game will not load by that stage.

Nobody is going to enable trace logging, NeoForge developer or not. It's also not helpful that the log message is missing the event class in one of the cases.

> if it were a true error, you would expect all of our dispatches to be guarded with something to the effect of if (!bus.isInErrorState()) { bus.dispatchEvent(...); } which is not the case.

What do you mean here?

PR #282, opened 2025-05-09
Update en_us.json

Thunderrock424242:
fixing some things with the missing classes instead of has mods that are missing it should be missing classes that are not found

checking for other grammer mistakes


sciwhiz12:
Hello! I don't think we use that translation key anymore, particularly because FML is hardcoded to reject any `ModLoadingIssue` that uses any translation key that starts with `fml.` but doesn't start with `fml.modloadingissue`. (See https://github.com/neoforged/FancyModLoader/blob/27a4a74731c0d2cd579fcf94f55e96e28d988ab5/loader/src/main/java/net/neoforged/fml/ModLoadingIssue.java#L28-L31)

We should look through all the translation keys and clean up those that are unused.

Thanks for your contribution anyway!

PR #297, opened 2025-06-19
Backport to 1.21.1: Remove the need to specify a bus in @EventBusSubscriber

Matyrobbrt:
1.21.1 backport of #290

To remove the breaking change, this PR introduces a `BOTH` bus which is the new default bus. To further maintain current runtime behaviour, unless an EBS-annotated class mixes listeners for the game and mod buses, it will still be directly registered as a `Class`, to facilitate unregistering.


Matyrobbrt:
@neoforged/bots backport 1.21.4

PR #319, opened 2025-07-04
Inline mixin service into FML and avoid parsing every class to a ClassNode

lukebemish:
This PR (a) disables loading of mixin's built-in services; (b) implements the necessary services in FML instead, (c) avoids passing every class to mixin as a ClassNode, thus avoiding ClassNode-ifying a _lot_ of classes, and (d) exposes some debug logging for statistics about the number of classes parsed to ClassNodes for transformation.


loader/src/main/java/net/neoforged/fml/loading/mixin/FMLMixinLaunchPlugin.java
```diff
@@ -0,0 +1,179 @@
...
+
+        modList.getModFiles().stream()
+                .map(ModFileInfo::getFile)
+                .forEach(file -> {
+                    final String modId = file.getModInfos().getFirst().getModId();
+                    for (ModFileParser.MixinConfig potential : file.getMixinConfigs()) {
+                        var existingModId = configModIds.putIfAbsent(potential.config(), modId);
+                        if (existingModId != null && !existingModId.equals(modId)) {
+                            LOGGER.error("Mixin config {} is registered by multiple mods: {} and {}", potential.config(), existingModId, modId);
+                            throw new IllegalStateException("Mixin config " + potential.config() + " is registered by multiple mods: " + existingModId + " and " + modId);
```
lukebemish, line 59: What exactly was the old behaviour here? I think throwing is most sensible (if we can't _fix_ this somehow, that is...) but it might be good for this to be non-breaking, so that could happen later.

lukebemish:
Marking this as ready for review. Of particular note, I could use feedback on the following:
- In general, should we plan on making the changes to make this non-breaking, or should we plan on pushing it to the next BC window?
- What should the behaviour be if two mods add mixin configs with the same name? As far as I can tell, it was just... I'm not sure. Silently ignoring it or something? (See `DeferredMixinConfigRegistration` before). So, the current behaviour I have here of throwing an exception could be a BC (but that can be changed, obviously), but more generally this feels like something that _shouldn't break stuff to begin with_, but I suppose that's maybe a longer term fix.
- The existing classes in the fml mixin package, and the relevant method on LoadingModlist, are all apparently public API. I should be able to reimplement them in a sensible enough fashion if we want this to be non-breaking.

So basically, should this be breaking and pushed to the next BC window, or not?

loader/src/main/java/net/neoforged/fml/loading/mixin/FMLMixinLaunchPlugin.java
lukebemish: And another point -- since this is an issue we _might_ actually expect to pop up in prod (unlike, say, the exceptions thrown for "the FMLMixinLaunchPlugin just doesn't exist" or the like), should use the ModLoadingIssue setup instead?

loader/src/main/java/cpw/mods/modlauncher/util/ServiceLoaderUtils.java
```diff
@@ -39,6 +39,12 @@ public static <T> Stream<T> streamWithErrorHandling(ServiceLoader<T> sl, Consume
         }).filter(Objects::nonNull);
     }
 
+    private static boolean isNotMixinService(ServiceLoader.Provider<?> provider) {
+        var clazz = provider.type();
+        var packageName = clazz.getPackageName();
+        return !packageName.equals("org.spongepowered.asm.launch") && !packageName.startsWith("org.spongepowered.asm.launch.");
```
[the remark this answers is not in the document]
lukebemish: And _technically_ there's subpackages not claimed by mixin either. So... yeah, I think you've got the right of it that it's not worth it.

PR #320, opened 2025-07-05
Read module-info.class of mod files using ModuleDescriptor and don't scan packages if unnecessary

shartte:
This removes ASM as a SJH dependency since the only thing it was used for can be done with `ModuleDescriptor#read`.
This uses the tests introduced in #315 to validate the behavior doesn't change.

@Technici4n this partially reverts your optimization for one special case: when the module-info.class doesn't contain a package list (it usually should if no special tooling was used to build it), the package scanning will be invoked synchronously.
If the module-info.class does contain a package list, we don't perform package scanning at all.

### DIFF
--- a/securejarhandler/src/main/java/cpw/mods/jarhandling/impl/ModuleJarMetadata.java
+++ b/securejarhandler/src/main/java/cpw/mods/jarhandling/impl/ModuleJarMetadata.java
@@ -20,10 +20,13 @@ import org.jetbrains.annotations.Nullable;
  */
 public class ModuleJarMetadata extends LazyJarMetadata {
     private final ModuleDescriptor originalDescriptor;
+    // If null, the package list from originalDescriptor will be used as-is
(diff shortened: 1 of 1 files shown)


marchermans:
> Thought about it, but name and version are used pretty much immediately as far as I know.

Sadface. But okey.

marchermans:
> I think reading the descriptor is cheap, scanning for packages is an order of magnitude more expensive... I can't think of a tool or use-case where the build tool wouldn't embed the package list though... maybe when in dev before packaging it into a jar?

Unsure.... It should work when it is a directory
But I am completely unsure.

Something we would need to test I think.

normally the compiler should generate the module-info.class file regardless of the end result, which should contain the package list.........

Keyword: Should.

Technici4n:
We could pass an empty package set to ModuleDescriptor.read and then add the packages later.

PR #325, opened 2025-07-09
Align package scanning more with how the JDK does it.

shartte:
Which results in the following:

- Packages that contain resources are picked up
- Packages that are not valid Java package names are ignored (every path segment needs to be a valid Java identifier).

This fixes issues with creating module metadata for LWJGL native modules, where the module-info.class has no package list, but exports a package called `windows.x64.[...]` which only contains the `.dll` but no `.class` file.


Technici4n (approved):
Yes, we should align with the JDK's scanning.

PR #326, opened 2025-07-09
Make `ModConfig.getFullPath` nullable

### LINKED ISSUE #236: Make `ModConfig.getFullPath` nullable
`ModConfig.getFullPath` currently has a comment on it wondering whether it's worth removing.

https://github.com/neoforged/FancyModLoader/blob/05a7b8219228f247036e47be124d0eec933b43de/loader/src/main/java/net/neoforged/fml/config/ModConfig.java#L64-L65

I personally use it to read another file, adjacent to the main config file, that contains secrets. One doesn't want to put secrets in the main config, as they are synced to the client!

*However*, there currently isn't a way (AFAIK) to determine whether the config file is being synced from disk over the network. As a result, there's no way to tell if `getFullPath` will throw. On 1.20.1, one could [poke the config file directly](https://github.com/cc-tweaked/CC-Tweaked/blob/5ba7f99326a6c0c9364fe008ec8ac9fdea3782a1/projects/forge/src/main/java/dan200/computercraft/ComputerCraft.java#L99), but as of #178 that's no longer possible, and you have to [`try`/`catch` the error instead](https://github.com/cc-tweaked/CC-Tweaked/blob/d697c47b8023ade33dda2198e34adf187e4ea9fc/projects/forge/src/main/java/dan200/computercraft/ComputerCraft.java#L179-L184). Ugly!

It would be nice if we just changed `getFullPath` to be nullable, rather than throwing an exception.

SquidDev:
Closes #236. To quote from that issue:

I currently use `ModConfig.getFullPath` in order to read another file adjacent to the main config file that contains secrets (you don't want them in the main config file, as they get synced over the network).

However, there currently isn't a way to determine whether the config file is being synced from disk over the network. As a result, there's no way to tell if `getFullPath` will throw, and so you have to [try/catch the error instead](https://github.com/cc-tweaked/CC-Tweaked/blob/d697c47b8023ade33dda2198e34adf187e4ea9fc/projects/forge/src/main/java/dan200/computercraft/ComputerCraft.java#L179-L184).

With this change, one can remove the `try`/`catch`, which makes this pattern much nicer!

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/config/ModConfig.java
+++ b/loader/src/main/java/net/neoforged/fml/config/ModConfig.java
@@ -69,11 +69,7 @@ public final class ModConfig {
      */
     @Nullable
     public Path getFullPath() {
-        if (this.loadedConfig != null && loadedConfig.path() != null) {
-            return loadedConfig.path();
-        } else {
-            return null;
-        }
+        return loadedConfig != null ? loadedConfig.path() : null;
     }
 
     void setConfig(@Nullable LoadedConfig loadedConfig, Function<ModConfig, ModConfigEvent> eventConstructor) {
--- a/loader/src/main/java/net/neoforged/fml/config/ModConfig.java
+++ b/loader/src/main/java/net/neoforged/fml/config/ModConfig.java
@@ -61,12 +61,18 @@ public final class ModConfig {
         return loadedConfig;
     }
 
-    // TODO: remove from public API?
+    /**
+     * Get the file path of the currently loaded config.
+     *
+     * @return The config's path. Will return {@code null} if the config is not currently loaded, or was loaded
+     *         from memory rather than from a file.
+     */
+    @Nullable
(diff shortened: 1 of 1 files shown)


loader/src/main/java/net/neoforged/fml/config/ModConfig.java
```diff
@@ -61,12 +61,18 @@ public IConfigSpec.ILoadedConfig getLoadedConfig() {
...
-    // TODO: remove from public API?
+    /**
+     * Get the file path of the currently loaded config.
+     *
+     * @return The config's path. Will return {@code null} if the config is not currently loaded, or was loaded
+     *         from memory rather than from a file.
+     */
+    @Nullable
     public Path getFullPath() {
         if (this.loadedConfig != null && loadedConfig.path() != null) {
```
Technici4n, line 72: We can drop the path null check now.

PR #328, opened 2025-07-11
Refactor how asynchronous mod file scans are handled to fix thread-safety

shartte:
We have a race condition in how we access the mod file scanning future, since we don't copy it to a local variable first.

This simplifies how the asynchronous mod file scanning is handled. We already had a future, so we'll always query that futures result instead of trying to use a volatile variable swap and two additional fields to carry the futures result.

This also refactors the ModFile slightly to make the future an implementation detail and not something that is set externally.

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ModFile.java
+++ b/loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ModFile.java
@@ -38,6 +38,7 @@ import net.neoforged.neoforgespi.locating.ModFileInfoParser;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.jetbrains.annotations.ApiStatus;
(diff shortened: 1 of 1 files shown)


loader/src/main/java/net/neoforged/fml/loading/moddiscovery/ModFile.java
```diff
@@ -47,13 +49,11 @@ public class ModFile implements IModFile {
...
     private Map<String, Object> fileProperties;
     private List<IModLanguageLoader> loaders;
-    private Throwable scanError;
     private final SecureJar jar;
     private final Type modFileType;
     private final Manifest manifest;
     private IModFileInfo modFileInfo;
-    private ModFileScanData fileModFileScanData;
-    private volatile CompletableFuture<ModFileScanData> futureScanResult;
+    private CompletableFuture<ModFileScanData> futureScanResult;
```
Technici4n, line 56: Nullable?

Technici4n (approved):

PR #333, opened 2025-07-15
Add MC-independent error display

XFactHD:
This PR adds an MC-independent interactive error display to allow displaying loading errors without having to fully load Minecraft in a crippled state.

The error display (right window) looks almost identical to the existing loading error screen (left window):
grafik

### DIFF
--- a/earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/error/ErrorDisplayWindow.java
+++ b/earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/error/ErrorDisplayWindow.java
@@ -9,7 +9,7 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.BiFunction;
-import java.util.stream.Stream;
+import java.util.function.Function;
 import net.neoforged.fml.ModLoadingIssue;
 import net.neoforged.fml.earlydisplay.render.EarlyFramebuffer;
 import net.neoforged.fml.earlydisplay.render.ElementShader;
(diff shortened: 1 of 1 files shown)


loader/src/main/resources/lang/en_us.json
```diff
@@ -77,5 +77,9 @@
   "fml.modloadingissue.enumextender.argument.constant.count_mismatch": "Parameter count {0} does not match argument count {1} of constructor {2} for field {3} in enum {4}, specified by mod {100,modinfo,id}",
   "fml.modloadingissue.enumextender.argument.constant.invalid_char": "Invalid character {0} at parameter index {1} for field {2} in enum {3}, specified by mod {100,modinfo,id}",
   "fml.modloadingissue.enumextender.argument.constant.unsupported_type": "Unsupported immediate argument type {0} at parameter index {1} for field {2} in enum {3}, specified by mod {100,modinfo,id}",
-  "fml.modloadingissue.enumextender.duplicate": "Found duplicate field {0} for enum {1}, provided by mods {2} and {3}"
+  "fml.modloadingissue.enumextender.duplicate": "Found duplicate field {0} for enum {1}, provided by mods {2} and {3}",
+  "fml.button.open.log": "Open log file",
+  "fml.button.open.crashreport": "Open crash report",
+  "fml.button.open.mods.folder": "Open Mods Folder",
+  "fml.button.quit": "Quit game"
```
Greg-21, line 84: Would it be possible for NeoForge to reuse the exact same translation keys and corresponding text strings that are now being reintroduced here in FML?

As someone who has been translating various Minecraft mods for several years now (but you don't have to be a translator to see this coming), I can already predict with a fair degree of certainty that there will inevitably be cases where, due to simple oversight, the same button labels end up being translated differently across NeoForge and FML. It's one of those situations where following the good ol' DRY principle now could save everyone some unnecessary headaches later on.

XFactHD: The duplicated translation keys will be dealt with in the PR integrating this into NeoForge.

Greg-21: Ah, got it! I wasn't aware that was the plan, so I just wanted to point it out in case it hadn't been considered. Thank you for letting me know.

Since I've got your attention and you're already at it, would you mind using proper title case capitalization for the new button labels, just to keep things consistent with vanilla Minecraft? I'm not really a fan of title case myself, but it seems to be Mojang's style, so probably best to stick with that if possible.

XFactHD: That's a fair point, I'll adjust the translations

PR #337, opened 2025-08-03
Allow Production & Userdev to use a fully formed Minecraft jar alongside an unmodified NeoForge universal jar

shartte:
The current layout to launch is rather convoluted in that it:

- Requires a separate "resource jar" (client-extra) for both production and dev (production also has "server-extra")
- In Production, the Jar is assembled from a myriad of individual jars (srg, patches) and loads the universal jar separately
- In Development, we use a joined jar for the code (not just joined between client+server, also joined between minecraft+neoforge), but still have the resources separately

This PR adds support for an alternative launch mode (which is auto-detected):

- The Minecraft Jar can be a Jar that contains both the classes and resources, alongside a neoforge.mods.toml for Minecraft
- If such a Minecraft Jar is detected, NeoForge will just be loaded from its Universal Jar (in Userdev, this just happens naturally since it is on the classpath and is detected as a normal mod)


Technici4n (approved):
Very nice.

PR #341, opened 2025-08-22
allow EventBusSubscriber to register static final field to eventbus

vfyjxf:
just what ##340 want


Technici4n (changes requested):
I am against this, we don't need more annotation magic.

TelepathicGrunt:
I too echo technician's stance. Annotation magic has always been a major pain point in teaching people how to mod and for beginners. This would only worsen the situation

PR #342, opened 2025-08-24
Implement JarContents without relying on UnionFS

shartte:
## VFS API Implementation without UnionFS / LayeredZipFileSystemProvider

This changes the implementation of the VFS API introduced in https://github.com/neoforged/FancyModLoader/pull/332 to not
simply rely on UnionFS.

Instead, it makes JarContents a sealed interface with these implementations:

| JarContents Implementation | Description                                                     |
|----------------------------|-----------------------------------------------------------------|
| `CompositeJarContents`     | A composite of other JarContents, optionally with filters       |
| `JarFileContents`          | A JarContents based on a JarFile                                |
| `FolderJarContents`        | A JarContents based on a NIO Root Path, intended for FS folders |
| `EmptyJarContents`         | An empty JarContents implementation                             |

Our use cases map to these roughly as follows:

- Any normal Mod Jar will be represented by a plain `JarFileContents`.
- During development, a mods output folders will be represented by a `CompositeJarContents` of two `FolderJarContents`
  (for the classes and resources folders).
- The current Minecraft Jar shenanigans will continue to work with a filtered `CompositeJarContents` over multiple
  `JarFileContents`.
- A mod wanting to provide a fully in-memory Jar could still do so via JimFS and using the root path for a
  `FolderJarContents`.
  There is no built-in way to tie the lifecycle of the FS to the FolderJarContents, but that could be added if needed.

For a given path, JarContents can be obtained using `JarContents JarContents#ofPath(Path path)`, which will
return a `JarFileContents` or `FolderJarContents` depending on whether the path is a file or a folder.

Getting a composite JarContent with filtering works by using
`JarContents JarContents#ofFilteredPaths(Collection<FilteredPath> paths)`.

Each filtered path is represented by this record:
`record FilteredPath(Path path, @Nullable CompositeJarContents.PathFilter filter)`

## JarInJar Changes

To simplify JarInJar handling, it now extracts the nested Jars to a temporary directory and uses simple `JarFileContents` to represent them.

This removes the need to ensure that `LayeredZipFileSystemProvider` is provided on the boot module / system classpath, due to the peculiarities of custom NIO FS implementations. Instead, we can use the same codepath as we do for normal mod jars (which is JarFile based), and will also use the same ResourcePack approach as vanilla resource packs.

## Signed / Multi-Release Jars

Since `JarFile` is now used natively, the default Java mechanism are used. These had to be reimplementad manually before due to use of `ZipFileSystem`.

## What about PackResources?

On the Minecraft side, we'll use the sealed nature of the interface to switch between Path-based, JarFile-based and a custom PackResource implementation. For any plain mod jar, a plain JarFile based implementation of PackResources should be used. Only for dev-time composite JarContents will we rely on our own implementation.

See the NeoForge side of this change here: https://github.com/neoforged/NeoForge/pull/2595

## Removed / Changed APIs

- Various places exposed a `moduleName` method which is replaced by `IModFile#getId()`
- Resource access methods in `LoadingModList` are gone since they exposed slow access using URI and NIO Path
- Removed public access to SecureJar from IModFile, what should be used instead is `IModFile#getJarContents()`
- Removed `findResource` from IModFile, use `IModFile#getJarContents().getResource(...)` instead
- `FMLLoader#addAccessTransformer` now uses a String to specify the path of the AT within the modfile, instead of a NIO Path

### DIFF
--- a/securejarhandler/src/main/java/cpw/mods/jarhandling/impl/CompositeJarContents.java
+++ b/securejarhandler/src/main/java/cpw/mods/jarhandling/impl/CompositeJarContents.java
@@ -23,6 +23,10 @@ import java.util.jar.Manifest;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
 
+/**
+ * A composite {@link JarContents} implementation which combines the contents of other jar contents transparently,
+ * and optionally applies filtering.
+ */
 @ApiStatus.Internal
 public final class CompositeJarContents implements JarContents {
     private final JarContents[] delegates;
@@ -32,10 +36,25 @@ public final class CompositeJarContents implements JarContents {
 
     private volatile Optional<String> checksum;
 
+    /**
+     * Constructs a new composite jar contents without filtering.
+     *
+     * @see JarContents#ofFilteredPaths
+     * @see JarContents#ofPaths
+     */
     public CompositeJarContents(List<JarContents> delegates) {
         this(delegates, null);
     }
 
+    /**
+     * Constructs a composite jar contents object with optional filtering.
+     *
+     * @param delegates The delegates that are combined. The delegation is last to first, which means the content in later delegates takes priority.
+     * @param filters   Optional filters for each entry in {@code delegates}. If {@code null}, no filtering is applied at all.
+     *                  If the filter for a delegate is {@code null}, no filtering is applied to that delegate.
+     * @see JarContents#ofFilteredPaths
+     * @see JarContents#ofPaths
+     */
     public CompositeJarContents(List<JarContents> delegates, @Nullable List<@Nullable PathFilter> filters) {
         // These checks are to prevent suboptimal usage. JarContents.ofFilteredPaths is going to guard against this
         if (delegates.isEmpty()) {
--- a/securejarhandler/src/main/java/cpw/mods/jarhandling/impl/EmptyJarContents.java
+++ b/securejarhandler/src/main/java/cpw/mods/jarhandling/impl/EmptyJarContents.java
@@ -11,11 +11,19 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.jar.Manifest;
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
 
+/**
+ * An immutable, empty jar content, which is identified by a {@link Path}, which may or may not exist.
+ */
+@ApiStatus.Internal
 public final class EmptyJarContents implements JarContents {
     private final Path path;
 
+    /**
+     * @see JarContents#empty(Path)
+     */
     public EmptyJarContents(Path path) {
         this.path = Objects.requireNonNull(path, "path");
     }
--- a/loader/src/test/java/net/neoforged/fml/loading/LauncherTest.java
+++ b/loader/src/test/java/net/neoforged/fml/loading/LauncherTest.java
@@ -39,7 +39,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
@@ -170,7 +169,6 @@ public abstract class LauncherTest {
 
         var previousCl = Thread.currentThread().getContextClassLoader();
         try (var cl = new URLClassLoader(urls, getClass().getClassLoader())) {
-            System.out.println("Constructed class-loader " + Objects.toIdentityString(cl));
             Thread.currentThread().setContextClassLoader(cl);
             return launchAndLoad(launchTarget);
         } finally {
--- a/securejarhandler/src/main/java/cpw/mods/jarhandling/JarContents.java
+++ b/securejarhandler/src/main/java/cpw/mods/jarhandling/JarContents.java
@@ -32,7 +32,7 @@ import org.jetbrains.annotations.Nullable;
  * for example.
  */
 @ApiStatus.NonExtendable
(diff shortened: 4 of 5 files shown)


loader/src/main/java/net/neoforged/neoforgespi/locating/IModFile.java
```diff
@@ -64,12 +65,15 @@ static IModFile create(SecureJar jar, ModFileInfoParser parser, IModFile.Type ty
...
-     * @param pathName The string representation of the path to find the mod resource on.
-     * @return The {@link Path} that represents the requested resource.
+     * <p>For mod files containing mods this will correspond with the mod id of the first mod contained in this file.
+     * <p>For non-mod jar files, an approach to generating a unique id is using the same algorithm used by Java
+     * to determine a Java module name for a given Jar file, but this is not guaranteed.
      */
-    Path findResource(String... pathName);
+    String getId();
+
+    JarContents getContents();
```
covers1624, line 76: Should probably have javadoc even if it is a trivial getter.

securejarhandler/src/main/java/module-info.java
```diff
@@ -9,6 +9,7 @@
     requires jdk.unsupported;
     requires java.base;
     requires static org.jetbrains.annotations;
+    requires java.desktop;
```
marchermans, line 12: What am I missing here that it now requires java.desktop?

securejarhandler/src/main/java/cpw/mods/cl/ModuleClassLoader.java
```diff
@@ -201,10 +201,9 @@ protected byte[] getClassBytes(final ModuleReader reader, final ModuleReference
     private Class<?> readerToClass(final ModuleReader reader, final ModuleReference ref, final String name) {
         var bytes = maybeTransformClassBytes(getClassBytes(reader, ref, name), name, null);
         if (bytes.length == 0) return null;
-        var cname = name.replace('.', '/') + ".class";
         var modroot = this.resolvedRoots.get(ref.descriptor().name());
         ProtectionDomainHelper.tryDefinePackage(this, name, modroot.jar().getManifest(), t -> modroot.jar().getManifest().getAttributes(t), this::definePackage); // Packages are dirctories, and can't be signed, so use raw attributes instead of signed.
-        var cs = ProtectionDomainHelper.createCodeSource(toURL(ref.location()), modroot.jar().verifyAndGetSigners(cname, bytes));
+        var cs = ProtectionDomainHelper.createCodeSource(toURL(ref.location()), null);
```
covers1624, line 206: Why can't we specify the code signers anymore? iirc this is required to prevent anyone packing classes into your signed packages. Perhaps just the url is enough though?

securejarhandler/src/main/java/cpw/mods/jarhandling/impl/ModuleDescriptorFactory.java
```diff
@@ -45,75 +42,48 @@ public static void scanAutomaticModule(
...
-                        if (!pkg.isEmpty()) {
-                            packageNames.add(pkg);
-                        }
-                    }
-                    return FileVisitResult.CONTINUE;
+        Map<String, JarResource> serviceProviderFiles = new HashMap<>();
+        jar.visitContent((relativePath, resource) -> {
+            // Ignore all content in META-INF except for service files
+            if (relativePath.startsWith("META-INF/services/")) {
+                var lastSeparator = relativePath.lastIndexOf('/');
```
covers1624, line 51: Same as my other comment above regarding `FilenameUtils.getName`, etc.

covers1624 (changes requested):
Some minor nits.

loader/src/main/java/net/neoforged/fml/loading/FMLPaths.java
```diff
@@ -19,6 +19,8 @@
 
 public enum FMLPaths {
     GAMEDIR(),
+    CACHEDIR(".cache"),
+    JIJ_CACHEDIR(".cache/jij"),
```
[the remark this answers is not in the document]
marchermans: Perfect. Thanks.

securejarhandler/src/main/java/cpw/mods/jarhandling/impl/CompositeJarContents.java
```diff
@@ -0,0 +1,289 @@
...
+    private final PathFilter @Nullable [] filters;
+    private final List<Path> contentRoots;
+
+    private volatile Optional<String> checksum;
+
+    public CompositeJarContents(List<JarContents> delegates) {
+        this(delegates, null);
+    }
+
+    public CompositeJarContents(List<JarContents> delegates, @Nullable List<@Nullable PathFilter> filters) {
```
marchermans, line 39: To me it does not make sense that both the `List` and its contents are Nullable.
Please clarify in the JDOC why this is needed on this method.

securejarhandler/src/main/java/cpw/mods/jarhandling/impl/EmptyJarContents.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.jar.Manifest;
+import org.jetbrains.annotations.Nullable;
+
+public final class EmptyJarContents implements JarContents {
+    private final Path path;
+
+    public EmptyJarContents(Path path) {
```
marchermans, line 19: Why is this referencing a `Path` and not just a simple singleton?

marchermans (changes requested):

securejarhandler/src/main/java/cpw/mods/jarhandling/impl/CompositeJarContents.java
marchermans: Ah okey, this is really not directly obvious.... Some more documentation would have been nice.....

Matyrobbrt (approved):

PR #350, opened 2025-08-27
Remove UnionFS and Clean Up Unused Code / Tests

shartte:
Only relevant after https://github.com/neoforged/FancyModLoader/pull/342

Should not actually remove anything we use, but Lazy *might* have been used by someone. If it comes up, we'll add it again (just not in SJH it really doesn't belong there).

### DIFF
--- a/loader/src/main/java/net/neoforged/fml/loading/LaunchContext.java
+++ b/loader/src/main/java/net/neoforged/fml/loading/LaunchContext.java
@@ -47,17 +47,13 @@ final class LaunchContext implements ILaunchContext {
                 for (var resolvedModule : layer.configuration().modules()) {
                     resolvedModule.reference().location().ifPresent(moduleUri -> {
                         try {
-                            locatedPaths.add(unpackPath(Paths.get(moduleUri)));
+                            locatedPaths.add(Paths.get(moduleUri));
                         } catch (Exception ignored) {}
                     });
                 }
             });
         }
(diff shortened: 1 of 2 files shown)


loader/src/main/java/net/neoforged/fml/loading/LaunchContext.java
```diff
@@ -58,9 +57,6 @@ final class LaunchContext implements ILaunchContext {
     }
 
     private Path unpackPath(Path path) {
-        if (path.getFileSystem() instanceof UnionFileSystem unionFileSystem) {
```
Technici4n, line 61: Nit: inline this method?

securejarhandler/src/main/java/cpw/mods/cl/ModuleClassLoader.java
```diff
@@ -117,11 +115,15 @@ public ModuleClassLoader(final String name, final Configuration configuration, f
...
-                        .orElse(ClassLoader.getPlatformClassLoader());
+                for (var parentLayer : parentLayers) {
+                    if (parentLayer.configuration() == k.configuration()) {
+                        var loader = parentLayer.findLoader(k.name());
+                        if (loader != null) {
+                            return loader;
+                        }
+                    }
+                }
+                return fallbackClassLoader;
```
Technici4n, line 126: This was changed from `ClassLoader.getPlatformClassLoader()`. This is probably fine, but please double-check.

Technici4n (approved):
Two small nitpicks, otherwise looks great!

Technici4n (approved):

PR #352, opened 2025-09-01
Remove Powermock

shartte:
We are primarily using Mockito in our projects, Powermock was imported from MCL and is somewhat of an odd-one-out.
Especially since it's not actually used to mock, rather we only use it to grab private fields which is somewhat of an anti-pattern for tests. I've opted to make public what makes sense, since the visibility is actually controlled more by not exposing the objects we want to keep private in the first place.

### DIFF
--- a/loader/src/main/java/cpw/mods/modlauncher/LaunchPluginHandler.java
+++ b/loader/src/main/java/cpw/mods/modlauncher/LaunchPluginHandler.java
@@ -19,6 +19,7 @@ import static cpw.mods.modlauncher.LogMarkers.MODLAUNCHER;
 
 import cpw.mods.jarhandling.SecureJar;
 import cpw.mods.modlauncher.api.IEnvironment;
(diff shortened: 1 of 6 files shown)


loader/src/jmh/java/cpw/mods/modlauncher/benchmarks/TransformBenchmark.java
```diff
@@ -39,72 +30,59 @@
...
-            classBytes = bos.toByteArray();
-        } catch (IOException e) {
-            e.printStackTrace();
-        }
-        plugins.put("dummy1", new ILaunchPluginService() {
-            @Override
-            public String name() {
-                return "dummy1";
-            }
+        final LaunchPluginHandler lph = new LaunchPluginHandler(Stream.of(
```
Technici4n, line 43: I don't know how much you want to clean up or not here. Cause we should potentially rename `final LaunchPluginHandler lph` to `var launchPluginHandler`.

PR #353, opened 2025-09-02
Annotate mixin configs with compat level

lukebemish:
Version of neoforged/FancyModLoader#339 targeting FML 10.x, leaving the other as a backport; updated to work with changes to FML mixin integration.


loader/src/main/java/net/neoforged/fml/loading/mixin/FMLMixinService.java
```diff
@@ -182,6 +187,17 @@ public Collection<IContainerHandle> getMixinContainers() {
...
+            // Mixin doesn't close this stream, so we use something that doesn't hold
+            // OS resources.
+            return new ByteArrayInputStream(content);
+        }
+
         return Thread.currentThread().getContextClassLoader().getResourceAsStream(name);
     }
+
+    public void addMixinConfigContent(String config, byte[] resource) {
+        mixinConfigContents.put(config, resource);
```
lukebemish, line 201: If we're _already_ doing this sort of thing to get the resources early enough -- we could just use this as an opportunity to rename the configs anyways so that multiple mods can use the same (normal) config file name. So, like, just rename them as `[modid]old.config.name.json` and then use that new name to tell mixin about the config.

loader/src/main/java/net/neoforged/fml/loading/mixin/FMLMixinLaunchPlugin.java
```diff
@@ -88,62 +93,124 @@ public void initializeLaunch(ITransformerLoader transformerLoader, NamedPath[] s
...
+                            mixinConfig.config(),
+                            existingInfo.modFile)
+                            .withAffectedModFile(modFile));
+                    continue;
+                }
+
+                // Grab the actual mixin config content from the Jar, which also allows us to print a nicer error if it's missing
+                byte[] configContent;
+                try {
+                    configContent = modFile.getContents().readFile(mixinConfig.config());
```
lukebemish, line 147: I feel like we should do the `requiredMods` check first -- does the file _really_ need to exist if the required mods don't? Maybe not relevant with the existing loader impls; just seems a strange restriction to add.

loader/src/main/java/net/neoforged/fml/loading/mixin/FMLMixinLaunchPlugin.java
```diff
@@ -88,62 +93,124 @@ public void initializeLaunch(ITransformerLoader transformerLoader, NamedPath[] s
...
+                if (configContent == null) {
+                    ModLoader.addLoadingIssue(ModLoadingIssue.error(
+                            "fml.modloadingissue.mixin.missing_config",
+                            mixinConfig.config())
+                            .withAffectedModFile(modFile));
+                    continue;
+                }
+
+                if (mixinConfig.requiredMods().stream().allMatch(id -> modList.getModFileById(id) != null)) {
+                    service.addMixinConfigContent(mixinConfig.config(), configContent);
```
lukebemish, line 160: As mentioned on discord -- this logic means that a mixin config _must_ come from the mod file declaring it. This is... I mean, frankly, a pretty sensible requirement, right? _But_ it's new compared to the old mixin impl and if we're going to add it we should let multiple mods use the same name the way I described elsewhere.

PR #423, opened 2026-05-18
Improve error handling of mixins during mod construction

IThundxr:
Simple change that just catches mixin transformer errors, unwraps them and blames the correct mod for them.

Previously if Mod A loaded class B, which Mod B had a mixin into which failed to apply, the error message would blame Mod A for that as it's the one that loaded Class B.


Su5eD (approved):
LGTM

marchermans (approved):

PR #426, opened 2026-06-09
Ignore client enum extensions on dedicated server

skirazavr:
Ignore client-only enum extensions on dedicated servers.

Currently RuntimeEnumExtender attempts to load enum prototypes
for net/minecraft/client/* classes on dedicated servers.

This causes ClassNotFoundException for client-only extensible enums
such as net.minecraft.client.gui.Hud$HeartType.

The change filters client-side enum extensions when running on
Dist.DEDICATED_SERVER.


skirazavr:
No need to add this cause works fine with dividing EnumParams with common and client sides

PR #430, opened 2026-06-21
[26.3] ELS renderer abstraction

XFactHD:
This PR implements a renderer abstraction layer for the early loading screen and changes the handover to throw away the existing window instead of handing it over to vanilla. This allows the ELS to continue using a (relatively) simple OpenGL renderer before the vanilla game opens (likely to be replaced with a Vk renderer once vanilla goes primarily/exclusively Vk) without preventing the full game from using Vulkan and also keeps the responsibility of the full Vk setup (including whatever extensions mods might request) in the vanilla B3D code instead of moving it to the ELS once it switches to Vk.


earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/SimpleBufferBuilder.java
```diff
@@ -332,51 +246,58 @@ private void ensureSpace(int newBytes) {
      * <p>
      * Uploading the buffers finishes drawing and resets for the next buffer operation.
      * <p>
-     * This should not be called in conjunction with {@link #draw()}
      *
      * @return The number of indexes that were uploaded.
      */
-    public int finishAndUpload() {
-        if (!building) throw new IllegalStateException("Not building.");
+    @Nullable
```
RogueLogix, line 252: This isn't actually nullable?

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/ELSDrawCollector.java
```diff
@@ -0,0 +1,125 @@
...
+        this.scissorWidth = width;
+        this.scissorHeight = height;
+    }
+
+    public void disableScissor() {
+        this.scissorEnabled = false;
+    }
+
+    public void submitDraw(ELSRenderPipeline pipeline, @Nullable ELSTexture texture, @Nullable SimpleBufferBuilder.Result bufferResult) {
+        if (bufferResult != null) {
```
RogueLogix, line 58: related to previous "not actuall nullable", practically this arg isn't nullable either

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/ELSDrawCollector.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public void execute(String label, ELSTexture target, ThemeColor clearColor, ELSBuffer vertexBuffer, int screenWidth, int screenHeight) {
+        int maxIndices = this.draws.stream()
+                .filter(Draw::indexed)
+                .mapToInt(Draw::indexCount)
+                .max()
+                .orElse(0);
+        ELSBuffer indexBuffer = this.backend.getQuadAutoIndexBuffer(maxIndices);
+
+        ByteBuffer uboData = MemoryUtil.memAlloc(2 * 4);
```
RogueLogix: MemoryStack

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/ELSRenderPipeline.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.util.List;
+import net.neoforged.fml.earlydisplay.render.ElementShader;
+import org.jetbrains.annotations.Nullable;
+
+public record ELSRenderPipeline(
+        ElementShader shader,
+        VertexFormat vertexFormat,
+        VertexFormat.Mode vertexMode,
+        @Nullable String sampler,
```
RogueLogix: pedantic, but, "texture", plz. i know GLSL calls it a "sampler2D", _it's confusing and not helped by GL not properly having separate samplers_

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/opengl/GlRenderBackend.java
```diff
@@ -0,0 +1,297 @@
...
+        GL33C.glBindBuffer(GL33C.GL_COPY_WRITE_BUFFER, 0);
+    }
+
+    @Override
+    public GlBuffer getQuadAutoIndexBuffer(int indexCount) {
+        return this.quadAutoIndexBuffer.acquire(indexCount);
+    }
+
+    @Override
+    public GlRenderPass createRenderPass(String label, ELSTexture target, ThemeColor clearColor) {
```
RogueLogix, line 126: scissor isn't disabled or state reset?

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/opengl/GlProgram.java
```diff
@@ -0,0 +1,131 @@
...
+        PointerBuffer sourcePointers = PointerBuffer.allocateDirect(1);
+        sourcePointers.put(0, fragmentShaderSource);
+        GL33C.glShaderSource(fragmentShader, sourcePointers, new int[] { fragmentShaderSource.remaining() });
+        sourcePointers.put(0, vertexShaderSource);
+        GL33C.glShaderSource(vertexShader, sourcePointers, new int[] { vertexShaderSource.remaining() });
+
+        // Compile the vertex and fragment elementShader so that we can use them
+        GL33C.glCompileShader(vertexShader);
+        if (GL33C.glGetShaderi(vertexShader, GL33C.GL_COMPILE_STATUS) == GL33C.GL_FALSE) {
+            throw new IllegalStateException("VertexShader linkage failure. \n" + GL33C.glGetShaderInfoLog(vertexShader));
```
RogueLogix, line 84: the exceptions leak the GL objects, is that fine because this is non-recoverable?

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/SimpleBufferBuilder.java
XFactHD: It is, the method returns null when no vertices were added to the buffer. In practice this currently never happens in the ELS but I prefer this over throwing an exception for an empty buffer.

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/opengl/GlProgram.java
XFactHD: It should be fine due to being unrecoverable but I can change it to clean up properly if you prefer.

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/opengl/GlRenderBackend.java
XFactHD: I did indeed forget to disable scissor (I was probably saved by the last draw having scissor disabled).
What do you mean by "state reset"? Should I be disabling anything else beyond the scissor state?

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/opengl/GlProgram.java
RogueLogix: don't care, just wanted to make sure unrecoverable.

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/opengl/GlRenderBackend.java
RogueLogix: reset the state to (0, 0, width height), disabling does effectively the same thing

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/SimpleBufferBuilder.java
RogueLogix: i managed to read that as a "throws"...

RogueLogix (dismissed):

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/DisplayWindow.java
```diff
@@ -135,6 +138,22 @@ public String name() {
         return "fmlearlywindow";
     }
 
+    @Override
+    public boolean isSupportedEnvironment() {
+        try (MemoryStack _ = MemoryStack.stackPush()) {
+            long handle = switch (Platform.get()) {
+                case FREEBSD, MACOSX -> 0L; // RenderDoc does not support MacOS and FreeBSD
+                case LINUX -> DynamicLinkLoader.dlopen("librenderdoc.so", DynamicLinkLoader.RTLD_NOW | DynamicLinkLoader.RTLD_NOLOAD);
```
RogueLogix: Does this add a refcount to the handle that needs a close?

XFactHD: It does indeed, I should have read the docs more thoroughly

RogueLogix (dismissed):

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/LoadingScreenRenderer.java
```diff
@@ -40,114 +26,126 @@
...
             Supplier<String> minecraftVersion,
-            Supplier<String> neoForgeVersion) {
-        this.glfwWindow = glfwWindow;
+            Supplier<String> neoForgeVersion,
+            boolean setupAutoRender) {
+        super("FML Early Loading Screen", backend, theme, externalThemeDirectory, LAYOUT_WIDTH, LAYOUT_HEIGHT);
         this.minecraftVersion = minecraftVersion;
         this.neoForgeVersion = neoForgeVersion;
+        this.elements = loadElements();
+        this.backend.releaseContextOwnership();
```
RogueLogix, line 68: this seems like the wrong place for this
the fact that its implied that backend.get() in the super constructor will automatically acquire the context is technically correct but also feels very _weird_ to me

XFactHD: I've added an explicit context acquisition to the super constructor. Is that sufficient or do you want me to move this release call out of the `LoadingScreenRenderer` constructor?

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/error/ErrorDisplayWindow.java
```diff
@@ -53,13 +62,10 @@ final class ErrorDisplayWindow {
     private static final int LIST_ENTRY_X = 30;
     private static final int LIST_CONTENT_WIDTH = DISPLAY_WIDTH - (LIST_ENTRY_X * 2);
     private static final int SCROLL_SPEED = 10;
+    private static final ThemeColor CLEAR_COLOR = ThemeColor.ofArgb(0xFF000000);
```
RogueLogix: nit: `OPAQUE_BLACK`?

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/TextureFormat.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) NeoForged and contributors
+ * SPDX-License-Identifier: LGPL-2.1-only
+ */
+
+package net.neoforged.fml.earlydisplay.render.backend;
+
+public enum TextureFormat {
+    RGBA(4),
+    RED(1),
```
RogueLogix: nit: RGBA8_UNORM/R8_UNORM (or RED8_UNORM)

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/ELSBuffer.java
```diff
@@ -0,0 +1,32 @@
...
+
+    enum Usage {
+        MAP_READ,
+        MAP_WRITE,
+        HINT_CLIENT_STORAGE,
+        COPY_DST,
+        COPY_SRC,
+        VERTEX,
+        INDEX,
+        UNIFORM,
```
RogueLogix: nit: most of these aren't used, and for the purpose of the ELS aren't necessary
reduced complexity to remove the ones that aren't needed

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/ELSDrawCollector.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        try (ELSRenderPass renderPass = this.backend.createRenderPass(label, target, clearColor)) {
+            renderPass.setViewport(this.viewportX, this.viewportY, this.viewportWidth, this.viewportHeight);
+            renderPass.bindUniform(ElementShader.UNIFORM_SCREEN_SIZE, this.backend.screenSizeUbo);
+
+            for (Draw draw : this.draws) {
+                SimpleBufferBuilder.Result result = draw.bufferResult;
+
+                renderPass.bindPipeline(draw.pipeline);
```
RogueLogix: (looking across PRs) this should be set lazily
the VK backend will proactively yeet descriptor state if this is called, without an early out.
it probably should early out and not do this, _but it does_

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/ELSDrawCollector.java
```diff
@@ -0,0 +1,123 @@
...
+                renderPass.bindVertexBuffer(vertexBuffer.slice(result.vertexOffset(), result.vertexCount() * (long) result.format().stride));
+                renderPass.bindIndexBuffer(result.indexed() ? indexBuffer : null);
```
RogueLogix: bind this once and leave it, non-indexed draws dont care about it being bound

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/ELSDrawCollector.java
```diff
@@ -0,0 +1,123 @@
...
+
+        try (ELSRenderPass renderPass = this.backend.createRenderPass(label, target, clearColor)) {
+            renderPass.setViewport(this.viewportX, this.viewportY, this.viewportWidth, this.viewportHeight);
+            renderPass.bindUniform(ElementShader.UNIFORM_SCREEN_SIZE, this.backend.screenSizeUbo);
+
+            for (Draw draw : this.draws) {
+                SimpleBufferBuilder.Result result = draw.bufferResult;
+
+                renderPass.bindPipeline(draw.pipeline);
+                renderPass.bindVertexBuffer(vertexBuffer.slice(result.vertexOffset(), result.vertexCount() * (long) result.format().stride));
```
RogueLogix: bind the vertex buffer once and use an offset in the draw command (GL: first or baseVertex; VK/RP: firstVertex)

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/ELSDrawCollector.java
```diff
@@ -0,0 +1,123 @@
...
+                renderPass.bindIndexBuffer(result.indexed() ? indexBuffer : null);
+                renderPass.bindTexture(ElementShader.UNIFORM_SAMPLER0, draw.texture);
```
RogueLogix: do this lazily, once again the backend isn't lazy about this and its expensive for VK

if its usually not the same texture, then this is fine

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/opengl/GlRenderBackend.java
```diff
@@ -0,0 +1,306 @@
...
+        int wleft = (int) (windowFBWidth * 0.5f - scale * width);
+        int wtop = (int) (windowFBHeight * 0.5f - scale * height);
+        int wright = (int) (windowFBWidth * 0.5f + scale * width);
+        int wbottom = (int) (windowFBHeight * 0.5f + scale * height);
+
+        GlState.bindDrawFramebuffer(0);
+        GlState.bindReadFramebuffer(((GlTexture) texture).fbo());
+        GlState.clearColor(backgroundColor.r(), backgroundColor.g(), backgroundColor.b(), 1f);
+        GL33C.glClear(GL33C.GL_COLOR_BUFFER_BIT | GL33C.GL_DEPTH_BUFFER_BIT);
+        GL33C.glBlitFramebuffer(
```
RogueLogix: this will cash on intel haswell iGPUs if the window is minimized, and afaik that isn't checked for elsewhere?

driver bug that causes a segfault

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/VertexFormat.java
```diff
@@ -0,0 +1,68 @@
...
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    public enum Element {
+        POS("position", 2, 2 * 4),
+        TEX("uv", 2, 2 * 4),
+        COLOR("color", 4, 4);
```
RogueLogix: nit: using proper formats (ie: RG32_FLOAT, RGBA8_UNORM) here helps with knowing what these actually are, rather than needing to look at the backend impl to figure out the formats

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/LoadingScreenRenderer.java
RogueLogix: while redundant both constructors doing an acquire/release, or neither doing it, would seem more right to me. granted, this is mostly just a code smell thing and not a functional issue.

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/LoadingScreenRenderer.java
```diff
@@ -40,114 +26,126 @@
...
 
-        // Create GL resources
-        this.theme = MaterializedTheme.materialize(theme, externalThemeDirectory);
-        this.elements = loadElements();
+    private void renderAutomatic() {
+        if (!this.renderLock.tryAcquire()) {
+            return;
+        }
+        try {
+            this.backend.acquireContextOwnership(false);
```
RogueLogix: nit (because this is off thread): ideally the context would be held by the secondary thread the entire time its rendering, and only released when handing off (or becoming error?)

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/RenderContext.java
```diff
@@ -142,36 +140,41 @@ public void renderProgressBar(Bounds barBounds, float fillFactor, int foreground
...
-    public void scissorBox(int x, int y, int width, int height) {
+    public void enableScissor(int x, int y, int width, int height) {
         // glScissor applies to the whole window, not just the viewport set via glViewport
-        GlState.scissorBox(
+        this.collector.enableScissor(
                 (int) (viewportOffsetX + x * viewportScale),
-                (int) (viewportOffsetY + y * viewportScale),
-                (int) (width * viewportScale),
-                (int) (height * viewportScale));
+                (int) (windowHeight - viewportOffsetY - (y + height) * viewportScale),
```
RogueLogix: why is windowHeight needed but not windowWidth?

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/SimpleBufferBuilder.java
```diff
@@ -332,51 +246,58 @@ private void ensureSpace(int newBytes) {
...
-                VERTEX_BUFFER_LENGTHS[format.ordinal()] = newVBOSize;
+                ELSBuffer oldBuffer = this.gpuBuffer;
+                this.gpuBuffer = backend.createBuffer(this.label, BUFFER_USAGE, newVBOSize);
+                if (oldBuffer != null) {
+                    backend.copyBufferToBuffer(oldBuffer.slice(), this.gpuBuffer.slice(0, oldBuffer.size()));
+                    oldBuffer.close();
+                }
             }
-            glBufferSubData(GL_ARRAY_BUFFER, 0, buffer);
+            backend.writeToBuffer(this.gpuBuffer.slice(this.gpuBufferOffset, this.index), this.buffer);
```
RogueLogix: nit: this call isn't free (VK stalls the GPU a lot), less of them is better

would also mean the above resize logic could be simplified by just using `memRealloc` for the native buffer and doing GPU buffer resize/upload once later

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/Texture.java
```diff
@@ -46,12 +24,20 @@ public int height() {
...
+    public int physicalWidth() {
+        return this.texture.width();
+    }
+
+    public int physicalHeight() {
+        return this.texture.height();
+    }
+
     /**
      * Loads a resource into an OpenGL texture.
```
RogueLogix: very nit: stale reference to OpenGL, also theres multiple but i'll leave just the one comment about it

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/backend/VertexFormat.java
```diff
@@ -0,0 +1,68 @@
...
+
+package net.neoforged.fml.earlydisplay.render.backend;
+
+import java.util.Arrays;
+
+public enum VertexFormat {
+    POS(Element.POS),
+    POS_TEX(Element.POS, Element.TEX),
+    POS_COLOR(Element.POS, Element.COLOR),
+    POS_TEX_COLOR(Element.POS, Element.TEX, Element.COLOR);
```
RogueLogix: nit: only one of these actually gets used?

earlydisplay/src/main/resources/net/neoforged/fml/earlydisplay/theme/gui.vert
```diff
@@ -1,14 +1,18 @@
...
-out vec2 fTex;
-out vec4 fColour;
+layout(std140) uniform screenSize {
+    vec2 screenSizeVec;
+};
+
+layout(location = 0) in vec2 position;
+layout(location = 1) in vec2 uv;
+layout(location = 2) in vec4 color;
+layout(location = 0) out vec2 fTex;
```
RogueLogix: not an issue with this review, but just something im going to note because these shaders get passed directly to OpenGL

Despite what GL_ARB_separate_shader_objects would lead you to believe, macOS' driver is broken and you must still use the same names in both stages. Probably best to leave a comment here about that because its easy to accidentally mess that up because every other platform implements the extension correctly and it works as expected.

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/DisplayWindow.java
```diff
@@ -388,72 +354,71 @@ public void initWindow() {
...
-        glfwShowWindow(window);
-        getLastGlfwError().ifPresent(error -> LOGGER.warn("Failed to show and position window: {}", error));
-        glfwPollEvents();
-    }
-
-    private void setWindowIcon() {
-        if (glfwGetPlatform() == GLFW_PLATFORM_COCOA) {
-            setMacosApplicationIcon();
-            return;
+        //if (!SDLVideo.SDL_ShowWindow(this.window)) {
```
RogueLogix: commented out code

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/DisplayWindow.java
```diff
@@ -464,31 +429,70 @@ private void winResize(long window, int width, int height) {
...
+        boolean maximized = (SDLVideo.SDL_GetWindowFlags(this.window) & SDLVideo.SDL_WINDOW_MAXIMIZED) != 0;
+        boolean fullscreen = (SDLVideo.SDL_GetWindowFlags(this.window) & SDLVideo.SDL_WINDOW_FULLSCREEN) != 0;
+        WindowState windowState = new WindowState(windowX, windowY, winWidth, winHeight, posValid, this.iconified, maximized, fullscreen);
+
+        if (destroyWindow) {
+            SDLVideo.SDL_DestroyWindow(this.window);
+            SDLInit.SDL_QuitSubSystem(SDLInit.SDL_INIT_VIDEO);
+        }
+
+        // Perform renderer re-init on the calling thread because the incoming B3D backend cannot move the context to another thread
```
RogueLogix: very nit: RenderPearl

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/DisplayWindow.java
```diff
@@ -464,31 +429,70 @@ private void winResize(long window, int width, int height) {
...
         }
 
-        return Optional.empty();
+        boolean maximized = (SDLVideo.SDL_GetWindowFlags(this.window) & SDLVideo.SDL_WINDOW_MAXIMIZED) != 0;
+        boolean fullscreen = (SDLVideo.SDL_GetWindowFlags(this.window) & SDLVideo.SDL_WINDOW_FULLSCREEN) != 0;
+        WindowState windowState = new WindowState(windowX, windowY, winWidth, winHeight, posValid, this.iconified, maximized, fullscreen);
+
+        if (destroyWindow) {
+            SDLVideo.SDL_DestroyWindow(this.window);
+            SDLInit.SDL_QuitSubSystem(SDLInit.SDL_INIT_VIDEO);
```
RogueLogix: curious why you need to do this, but Vanilla doesn't?

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/DisplayWindow.java
```diff
@@ -515,33 +520,36 @@ public long takeOverGlfwWindow() {
...
         completeProgress();
 
-        glfwMakeContextCurrent(window);
-        // Set the title to what the game wants
-        glfwSwapInterval(0);
-        // Clean up our hooks
-        glfwSetWindowSizeCallback(window, null).close();
-        this.repaintTick = renderer::renderToScreen; // the repaint will continue to be called until the overlay takes over
-        return window;
+        return renderer.getBackend();
```
RogueLogix: code smell: the backend gets returned, but sometimes is destroyed first. IMO always return a valid one and its on the caller to destroy it if they want to (or dont want to).

or, return null if it was destroyed?

earlydisplay/src/test/java/net/neoforged/fml/earlydisplay/render/SimpleFontTest.java
```diff
@@ -3,11 +3,13 @@
  * SPDX-License-Identifier: LGPL-2.1-only
  */
 
+/*
```
RogueLogix: commented out entire file

earlydisplay/src/test/java/net/neoforged/fml/earlydisplay/render/WithOffScreenGLSurface.java
```diff
@@ -3,88 +3,75 @@
  * SPDX-License-Identifier: LGPL-2.1-only
  */
 
+/*
```
RogueLogix: two of them!

loader/src/main/java/net/neoforged/fml/loading/EarlyLoadingScreenController.java
```diff
@@ -19,18 +21,20 @@ static EarlyLoadingScreenController current() {
...
-     */
-    long takeOverGlfwWindow();
+    /// Stop polling SDL events prior to vanilla window creation.
+    void stopEventPolling();
+
+    /// Hands over control of the early loading screen to Minecraft, destroys the ELS' own window
+    /// and restarts its rendering with the provided backend on the window spawned by vanilla.
+    ///
+    /// This method can only be called once and once this method is called, any off-thread
+    /// interaction with the window seizes.
```
RogueLogix: stale comment because its now two windows and not "the window"

loader/src/main/java/net/neoforged/fml/loading/EarlyLoadingScreenController.java
```diff
@@ -19,18 +21,20 @@ static EarlyLoadingScreenController current() {
...
+    ///
+    /// @return the state of the ELS window to be applied to the vanilla window
+    WindowState handOverToMinecraft(Supplier<Object> renderBackend);
 
     /**
-     * After calling {@linkplain #takeOverGlfwWindow() taking over} the main window, the game may still want to
+     * After calling {@linkplain #handOverToMinecraft(Supplier) taking over} the main window, the game may still want to
```
RogueLogix: comment was broken, but also stale now

loader/src/main/java/net/neoforged/fml/startup/FatalErrorReporting.java
```diff
@@ -132,16 +132,11 @@ public static void reportFatalError(String message) {
         if (!GraphicsEnvironment.isHeadless()) {
             showErrorUsingSwing(message);
         } else {
+            // TODO: check whether this is still necessary
```
RogueLogix: have you checked?

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/LoadingScreenRenderer.java
XFactHD: I've adjusted the three constructors to all explicitly acquire the context and the `AbstractEarlyScreen` and `LoadingScreenRenderer` constructors to also release it. The `ErrorDisplayWindow` constructor doesn't release it because it would just be re-aquiring it in `ErrorDisplay` right after the constructor call.

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/LoadingScreenRenderer.java
XFactHD: I can't see any reasonable way of releasing the context when the handoff happens and without that the handoff (which happens on a different thread) would fail to acquire the context, so there's no way around acquiring and releasing the context every frame.

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/RenderContext.java
XFactHD: I need to flip the scissor box vertically the same way vanilla does in `GuiRenderer#enableScissor()` because *coordinate systems*

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/DisplayWindow.java
XFactHD: Vanilla uses `SDL_Quit()` instead which forcefully kills all subsystems. I can't afford terminating everything though, so I have to use the "soft" option of relinquishing my reference to the subsystem this way.

earlydisplay/src/test/java/net/neoforged/fml/earlydisplay/render/SimpleFontTest.java
XFactHD: The test and the associated `WithOffScreenGLSurface` JUnit extension are currently broken due to SDL and the software GL renderer not getting along and I can't exactly use lavapipe for GL

loader/src/main/java/net/neoforged/fml/startup/FatalErrorReporting.java
XFactHD: I have done a quick test on Windows, but I'm not sure which platforms this even applied to, so I'm not comfortable taking this as conclusive evidence.

PR #435, opened 2026-07-31
Fix failure to recompute frames in mixins involving mixin-generated classes in stack frames

### LINKED ISSUE #434: Mixin fails to load generated class
## The issue

There is a classloading error that happens when the mixin class writer attemps to read a generated class, when calculating the mixin class hierarchy for writing. The FML method for this operation, `upToFrames`, does not support loading the bytecode of generated classes and will lead to a CNFE, which then crashes the game.

## Example code

The Mixin code needed to reproduce this is rather specific as the frame structure must force the class writer try to access the anonymous mixin class. Below is an example Mixin I used to test this issue. It is a stripped down version of the original from the Combatify mod.

```java
@Mixin(Player.class)
public class PlayerMixin {
    @Inject(
        method = "attack",
        at = @At(
            value = "INVOKE",
            target = "Lnet/minecraft/world/entity/Entity;getDeltaMovement()Lnet/minecraft/world/phys/Vec3;"
        )
    )
    public void injectCrit(Entity target, CallbackInfo ci, @Local(ordinal = 3) LocalFloatRef combinedDamage) {
        runCrit(!isStateVanilla() ? combinedDamage : new LocalFloatRef() {
            public float get() {
                return 0;
            }

            public void set(float v) {
            }
        });
    }

    private boolean runCrit(LocalFloatRef ref) {
        return true;
    }

    private boolean isStateVanilla() {
        return false;
    }
}
```

When computing frames, `TransformerClassWriter#computeHierarchyFromFile` will call `ClassHierarchyRecomputationContext#upToFrames` to retrieve raw bytecode for the anonymous mixin class created for `LocalFloatRef` inside `PlayerMixin`. However, the method does not support this and will end up throwing `ClassNotFoundException`, crashing the whole process. The game will not boot beyond the loading screen.

## Steps to reproduce

1. Add the mixin from above into the game (e.g. using a mod)
2. Launch the game
3. During the loading process, an error screen will be displayed and loading will stop

You can place a conditional breakpoint inside `upToFrames` in `TransformingClassLoader.java` to trigger when the class name contains `Anonymous` and inspect the call tree like that.

## Expected behavior

The anonymous class is generated by Mixin and loadable by the classloader from the usual context. Ideally, the `upToFrames` method should be modified to support returning the bytecode of generated classes, so that the class hierarchy can be computed correctly, as it's for other non-generated classes.

## Additional information

Tested on NeoForge 26.1.2.84 in userdev. The `mixin.debug.export` was turned off - the error will still occur with it enabled, but the stacktrace might differ.

Error log: https://gist.github.com/Su5eD/9ed8b3bf879375808c929ba7648e477a

lukebemish:
Fixes #434

In certain scenarios frame recomputation of Mixin classes needs to see the superclass of a class mixin generates. However, mixin-generated classes (as well as normal mixin classes) themselves may need frame computation, so that transform cannot be placed before `neoforge:computing_frames`! (Basically, a given class processor can _either_ have the ability to have its changes be counted for frame computing, or can request computing itself. This is to avoid circularity).

I fix this through a few changes:
- Make a new class processor, `neoforge:mixin_frame_context`, responsible for making any mixin changes that will affect frame computation
- Move the bytecode capture point used by mixin to look up class hierarchy to this point (this is necessary to avoid circularity)
- `neoforge:mixin_frame_context` runs `generateClass` on classes-to-be-generated on a copy of the received class node, then copies over any changes in super type to the actual classnode.This occurs _before_ `neoforge:computing_frames`
- `neoforge:mixin` runs `generateClass` to properly generate the class normally, later, as before.

Basically: I apply the changes in super type earlier by re-running the generator. Generators are safe to re-run -- FML already did that in `FMLClassBytecodeProvider`. The capture point has to be moved earlier because generators could request stuff from the mixin service, including looking up class info; I suspect this will not have any interesting effects in practice.

### DIFF
--- a/loader/src/test/java/net/neoforged/fml/loading/MixinConfigTest.java
+++ b/loader/src/test/java/net/neoforged/fml/loading/MixinConfigTest.java
@@ -206,6 +206,62 @@ public class MixinConfigTest extends LauncherTest {
         assertThat(Mixins.getConfigs()).extracting("name").containsOnly("test.mixins.json");
     }
 
+    @Test
+    void testFrameRecomputationWithGeneratedClasses() throws Exception {
+        installation.setupProductionClient();
+        installation.buildModJar("mixin-test.jar")
+                .withTestmodModsToml(modsToml -> modsToml.addMixinConfig("test.mixins.json").addMod("test"))
+                .addTextFile("test.mixins.json", """
+                        {
+                            "package": "test.mixin",
+                            "mixins": ["RequiresFrameRecompute"]
+                        }
+                        """)
+                .addClass("test.target.Super", """
+                        public class Super {}
+                        """)
+                .addClass("test.target.Target", """
+                        import net.neoforged.fml.common.Mod;
+
+                        @Mod("test")
+                        public class Target {
+                            {
+                                test(null);
+                            }
+
+                            public static void test(Super obj) {}
+                        }
+                        """)
+                .addClass("test.mixin.RequiresFrameRecompute", """
+                        import org.spongepowered.asm.mixin.injection.At;
+                        import org.spongepowered.asm.mixin.injection.Inject;
+                        import org.spongepowered.asm.mixin.Mixin;
+                        import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+                        import test.target.Super;
+
+                        @Mixin(test.target.Target.class)
+                        public class RequiresFrameRecompute {
+                            @Inject(
+                                method = "test",
+                                at = @At("HEAD")
+                            )
+                            private static void testInject(Super obj, CallbackInfo ci) {
+                                // makes recomputing the stack necessary
+                                takesLocalType(alwaysFalse() ? obj : new Super() {});
+                            }
+
+                            private static boolean alwaysFalse() {
+                                return false;
+                            }
+
+                            private static void takesLocalType(Super object) {}
+                        }
+                        """)
+                .build();
+
+        var result = launchAndLoad("neoforgeclient");
+    }
+
     /**
      * Tests that Mixin configs declared only via the manifest are correctly picked up by Mixin.
      * <p>This is used by mixinextras, for example.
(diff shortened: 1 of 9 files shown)


lukebemish:
This PR also includes a test replicating #434; this test will fail without the changes from this PR.

Su5eD (dismissed):

lukebemish:
Is this PR waiting on anything on my end?

Su5eD:
@Matyrobbrt dew it

loader/src/test/java/net/neoforged/fml/loading/MixinConfigTest.java
```diff
@@ -206,6 +206,62 @@ void testRequestedMixinBehaviorIsValid() throws Exception {
         assertThat(Mixins.getConfigs()).extracting("name").containsOnly("test.mixins.json");
     }
 
+    @Test
+    void testFrameRecomputationWithGeneratedClasses() throws Exception {
```
sciwhiz12: Can this test be moved to a separate test class? `MixinConfigTest` seems to be primarily concerned with checking Mixin configs are detected properly, and I'd not want to muddle it with a test outside of that scope.

sciwhiz12 (commented):
I tried to understand the code, but it eludes my understanding, so I'm relying on the above approving review and the test (which I checked to fail without the fix, and succeeds with the fix).

loader/src/test/java/net/neoforged/fml/loading/MixinConfigTest.java
lukebemish: `MixinConfigTest` contains some useful utilities necessary for getting tests involving mixin to work _at all_ -- namely, the nonsense that is `@BeforeEach void cleanUpMixins()`. Any test involving mixins has to be in a test class with that `@BeforeEach` -- I stuck it here to avoid duplication but I can try and pull that out into a separate utility or something (would sticking it in a superclass work? I can check).

sciwhiz12: Alternatively, we could rename this class `MixinTest` then.

sciwhiz12 (approved):

PR #438, opened 2026-08-08
Fix loader warnings for discouraged mods

### LINKED ISSUE #268: in 1.21.1 Minecraft, using incompatible clause shows missing translation for "No Reason"
Was using latest neoforge for that version and discovered this in prod
Image

Thodor12:
Currently, when you have a loader warning for a discouraged mod, all you see is this:

Warning 1

This is not useful at all, the actual problem is only mentioned in the logs, but the logs also don't state the reason field we provided in the mods.toml. So the actual reason field is nowhere visible at all.

This PR intends to fix this, by making sure the loader warning messages are correctly dumped in the output again, and the correct "No reason" translation strings are used again.

This is the result:

Warning 2

I also added 3 unit tests that now verify that the loader correctly outputs these warnings when an incompatible/discouraged mod is found, so this can't secretly break again.

I made the initial fix for 1.21.1, with the exact same code added, so it should be possible to easily backport this all the way to 1.21.1


sciwhiz12 (approved):
It always slightly pains me to see that subtle difference of `fml.modloadingissue` versus `fml.modloading`.

Code looks good, tests work (fail without the fix, succeed with the fix).

Thodor12:
@sciwhiz12 Can this be backported to 1.21?

PR #439, opened 2026-08-09
Fix JVM crash on Mac OS when resizing the window

mezz:
I was testing resizing the gui during early loading and got the JVM to crash easily with an hs_err.

We need to call `glfwPollEvents()` from within the loading screen render lock to prevent the crash during resize on Mac OS.


XFactHD (changes requested):
I would prefer folding this into the existing PR.

mezz:
@XFactHD I created a new PR against yours here: https://github.com/XFactHD/FancyModLoader/pull/1
you can merge it to fold this into your existing PR

PR #440, opened 2026-08-09
Fix log spam and possible hang if you resize the window very small on Mac OS

mezz:
This prevents progress-bars from collapsing to zero pixels and producing repeated stencil errors while the window is resized to a very small size.

### DIFF
--- a/earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/RenderContext.java
+++ b/earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/RenderContext.java
@@ -142,12 +142,16 @@ public record RenderContext(
 
         blitTexture(sprites.progressBarBackground(), barBounds);
 
+        if (fillFactor == 0) {
+            return;
+        }
+
         GlState.scissorTest(true);
         scissorBox(
-                (int) barBounds.left(),
-                (int) barBounds.top(),
-                (int) (barBounds.width() * fillFactor),
-                (int) barBounds.height());
+                barBounds.left(),
+                barBounds.top(),
+                barBounds.width() * fillFactor,
+                barBounds.height());
         blitTexture(sprites.progressBarForeground(), barBounds, foregroundColor);
         GlState.scissorTest(false);
     }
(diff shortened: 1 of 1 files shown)


earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/RenderContext.java
```diff
@@ -167,11 +171,19 @@ public void fillRect(float x, float y, float width, float height, int colorTop,
     }
 
     public void scissorBox(int x, int y, int width, int height) {
+        scissorBox((float) x, (float) y, (float) width, (float) height);
+    }
+
+    private void scissorBox(float x, float y, float width, float height) {
+        if (!(width > 0) || !(height > 0)) {
```
shartte: The conditional is a bit weird, I'd have expected  <= 0 but oh well

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/RenderContext.java
```diff
@@ -142,12 +142,16 @@ public void renderProgressBar(Bounds barBounds, float fillFactor, int foreground
 
         blitTexture(sprites.progressBarBackground(), barBounds);
 
+        if (fillFactor == 0) {
+            return;
+        }
```
shartte, line 147: Isn't this the actual fix? :-P

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/RenderContext.java
XFactHD: I'm pretty sure this is related to floating point precision issues in the same way that MC's `Mth.equal()` with epsilon is.

XFactHD (changes requested):
I would prefer folding this into the existing PR, particularly because I'm already doing part of this fix (throwing an exception for non-positive scissor sizes) in that PR.

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/RenderContext.java
mezz: ah sorry, earlier instead of throwing I had wrapped everything in `if (width > 0 && height > 0) {` 
I did some invert-condition action in intellij and then didn't pay attention to the weird result

earlydisplay/src/main/java/net/neoforged/fml/earlydisplay/render/RenderContext.java
mezz: hmm maybe yeah. I was throwing the kitchen sink at this thing so I may have done more than necessary. I'll test it out after work

mezz:
I was able to remove the complicated bits and still keep it working, please take another look :)

mezz:
Folding this into the existing PR here: https://github.com/XFactHD/FancyModLoader/pull/3

PR #441, opened 2026-08-09
Set the macOS application icon

mezz:
GLFW cannot set regular window icons on macOS, so the early-loading process was appearing with a generic Java or terminal icon in the Dock. Now it'll show the neoforge icon like the other platforms.


shartte (approved):
Can't test this, but code looks good to me

PR #442, opened 2026-08-11
Build: Update to Mixin 0.17.3

LlamaLad7:
Contains various bug fixes: https://github.com/FabricMC/Mixin/compare/0.17.1%2Bmixin.0.8.7...0.17.3%2Bmixin.0.8.7


sciwhiz12 (changes requested):
Oops, we forgot about this. Could you update this to 0.17.4+mixin.0.8.7?

If updating that version in the `MixinConfigTest` is required whenever we update the Mixin version, we should (in a separate PR) look into making that automatically updated by Renovate.

